Issue #43, opened 2018-07-19
Implement more fine grained tagging in prometheus metrics


mshutt:
**Describe the feature you would like and why you want it**

Add additional tags to prometheus check value metrics:

https://github.com/Comcast/kuberhealthy/blob/6dab3697e6626f1e76f3e3a2f3e4941524046f80/pkg/metrics/exporter.go#L40

e.g. check=$checkname,namespace=$namespace for additional grainularity.

**Additional context**

We're feeding these into DataDog such as:  https://photos.app.goo.gl/4o989E9uvSWkoVmd6

It would be excellent if we had per check/namespace tuples that we could filter on and use in alerts if a check value drops below 1.

integrii:
@ihoegen did the Prometheus integration. He is off in Mexico on a honeymoon right now though...  I assigned him to this one.

ihoegen:
The replace is there because for a little bit the name of the metric was the name of the check, and spaces made things tricky. Now it's just leftover code I could remove. I'll update it to parse that when I get back on Monday

mshutt:
Right, but we don't want to remove the fidelity were getting there since
the podstatuscheck can execute for N namespaces same as the podhealthcheck

Right?  Just create a tuple of tags {check=type,namespace=namespace}

ihoegen:
Yeah that makes sense. I'll remove the parsing of the spaces, and use the namespace field from the check object to populate that tuple

ihoegen:
#56 addresses this

PR #56, opened 2018-08-17
Exposes the namespace of the check to the CRD and the metrics endpoint

ihoegen:
Addresses #43

### DIFF
--- a/cmd/kuberhealthy/kuberhealthy.go
+++ b/cmd/kuberhealthy/kuberhealthy.go
@@ -12,6 +12,7 @@
 package main
 
 import (
+	"fmt"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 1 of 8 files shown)


ihoegen:
Yeah the Component Status Check returns "" for the CheckNamepsace method

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -390,3 +399,13 @@ func (k *Kuberhealthy) getCurrentState() (health.State, error) {
...
 }
+
+// getCheck returns a Kuberhealthy check object from its name, returns an error otherwise
+func (k *Kuberhealthy) getCheck(name string) (KuberhealthyCheck, error) {
+	for _, c := range k.Checks {
+		if c.Name() == name {
+			return c, nil
+		}
+	}
+	return nil, fmt.Errorf("Could not find Kuberhealthy check with name %s", name)
```
lolimjake: This project uses the logrus package for logging output.

integrii:
The alternative would be to alter the interface's return `[]string` here:

https://github.com/Comcast/kuberhealthy/blob/master/cmd/kuberhealthy/kuberhealthyCheck.go#L35

I'm not sure what is worse - having to implement dead methods, or having to assume some set of return values...

I think I lean toward the way you've done it...

cmd/kuberhealthy/kuberhealthy.go
ihoegen: fmt.Errorf doesn't output, it returns an error

PR #66, opened 2018-09-26
Go modules

integrii:
Implements go modules and removes an excessive logging line.  Adds a .gitignore so that the vendor directory and binary do not end up in the codebase by accident.


integrii:
hold off. i've done this wrong.

PR #70, opened 2018-10-17
Easy spec install

integrii:
I moved everything for grafana and helm into a `deploy` folder, then reworked the install instructions to be more concise and easy to do with less reading.


integrii:
opening a new request from the gh-pages branch...

PR #79, opened 2018-11-01
Fix a few typos identified from Go Report Card

iamjarvo:


integrii:
Thanks for fixing these and taking the time to make a PR!

Issue #80, opened 2018-11-15
Pod restart is throwing false alarms


integrii:
It appears that the pod restart checker is triggering on pods that have not restarted five times in the last hour.  This only happens once the pods _have_ actually restarted five times previously.  Likely the counter is not being purged correctly.

![image](https://user-images.githubusercontent.com/98695/48587752-3b04e280-e8e9-11e8-98c8-3783ec274114.png)
![image](https://user-images.githubusercontent.com/98695/48587758-3fc99680-e8e9-11e8-8ff0-b50e0d502109.png)

lolimjake:
The pod in the example was is part of a stateful set and had been restarted within the past hour (the timeframe the PRC looks for restarts over).  I believe the previous iteration of this pod WAS restarted more than 5 times in the last hour, but the pod was also terminated completely, wiping the restart count back to 0 when you check after the fact.  

I just tested and terminating a pod with a restart count > 5 generates an alert - thats where I think this bug is. I'll do a bit more investigation.

lolimjake:
The bug lies here: https://github.com/Comcast/kuberhealthy/blob/8f25870ae211caa721c6cd6ee2626a08882f2f88/pkg/checks/podRestarts/podRestarts.go#L207 .  

We are collecting metrics based on pod names - since stateful set pods come back with the same name, kuberhealthy treats the terminated and newly created stateful set pod as the same pod.  In the linked line, Kuberhealthy is only measuring the delta between the largest recorded value and the smallest recorded value of restarts.  Since the new stateful set pod of the same name is created with a restart count of 0, if the outgoing pod had a restart count of more than 5, Kuberhealthy alerts that the delta between the max and min restart values was greather than 5.

We can handle this by either comparing something more specific than pod names (resource version?) Or detecting when the restart count is reset to 0 and clearing that pod's history upon reset detection.

lolimjake:
https://github.com/Comcast/kuberhealthy/pull/89
https://github.com/Comcast/kuberhealthy/pull/90
https://github.com/Comcast/kuberhealthy/pull/92

Fixed this, wrote a proper test and It's been merged into unstable

PR #86, opened 2019-01-03
Prc statefulset fix

lolimjake:
This fix should compare the most recent restart observation to the previous restart observation and determine if it has dropped from a non zero value to zero, indicating that a stateful set pod was deleted and recreated.  

This fix should resolve this bug: https://github.com/Comcast/kuberhealthy/issues/80


lolimjake:
Missed merge into unstable vs master

PR #87, opened 2019-01-03
Prc statefulset fix

lolimjake:
This fix should compare the most recent restart observation to the previous restart observation and determine if it has dropped from a non zero value to zero, indicating that a stateful set pod was deleted and recreated.

This fix should resolve this bug: https://github.com/Comcast/kuberhealthy/issues/80


pkg/checks/podRestarts/podRestarts.go
```diff
@@ -171,6 +173,28 @@ func (prc *Checker) reapPodRestartChecks(currentPods *v1.PodList) {
 			if observation.Time.Before(time.Now().Add(-time.Hour)) {
 				prc.RestartObservations[podName] = deleteFromSlice(i, prc.RestartObservations[podName])
 			}
+			// Find the most recent PRC observation
+			if observation.Time.After(mostRecentObservation.Time) {
+				mostRecentObservation = observation
+			}
+			// Find the second most recent PRC observation
+			if observation.Time != time.isZero() && observation.Time != mostRecentObservation.Time && observation.Time.After(secondMostRecentObservation.Time) {
```
ihoegen: I'd consider using the Time.Equal function instead of ==

Also Nit: instead of making sure that `observation.Time != mostRecentObservation.Time`, it'd make sense to do `observation.Time.Before(mostRecentObservation.Time)`

Issue #16, opened 2018-06-21
Check all kube-dns pods functionality


integrii:
**Describe the solution you would like to see happen**
Sometimes specific kube-dns pods are not returning appropriate responses.  We could check them directly with DNS queries.

**How do you propose we build it?**
Make a new check.  Check lists kube-dns pods in kube-system.  DNS tests are executed on each for various internal service names.

ihoegen:
Ran into an issue on a cluster where kube-dns wasn't resolving, which would've been caught if we had this. Going to work on implementing

integrii:
Thanks Ian!

PR #88, opened 2019-01-03
#16 Add DNS status check

ihoegen:
Addresses the concerns in #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,18 @@ A command line flag exists `--podCheckNamespaces` which can optionally contain a
 - Error state toleration: 5 minutes
 - Check name: `podStatus`
 
+#### DNS
+
+Checks for failures with DNS, including resolving within the cluster and outside of the cluster. Default endpoints to resolve: kubernetes.default, aws.amazon.com, cloud.dns.com
+
+A command-line flag exists `--dnsEndpoints` which can optionally include a comma separated list of DNS endpoints to test. 
+
+- Timeout: 1 minutes
+- Check Interval: 2 minutes
+- Error state toleration: 5 minutes
+- Check name: `dnsStatus`
+
+
 ### Security Considerations
 
 By default, Kuberhealthy exposes an insecure (non-HTTPS) status endpoint without authentication. You should never expose this endpoint to the public internet. Exposing Kuberhealthy's status page to the public internet could result in private cluster information being exposed to the public internet when errors occur and are displayed on the page.
--- a/README.md
+++ b/README.md
@@ -156,8 +156,8 @@ Checks for failures with DNS, including resolving within the cluster and outside
 A command-line flag exists `--dnsEndpoints` which can optionally include a comma separated list of DNS endpoints to test. 
 
 - Timeout: 1 minutes
-- Check Interval: 2 minutes
-- Error state toleration: 5 minutes
+- Check Interval: 30 seconds
+- Error state toleration: 1 minute
 - Check name: `dnsStatus`
 
 
--- a/cmd/kuberhealthy/main.go
+++ b/cmd/kuberhealthy/main.go
@@ -36,7 +36,7 @@ import (
 var kubeConfigFile = filepath.Join(os.Getenv("HOME"), ".kube", "config")
 var listenAddress = ":8080"
 var podCheckNamespaces = "kube-system"
-var dnsEndpoints = ""
+var dnsEndpoints = []string{}
 
 // shutdown signal handling
 var sigChan chan os.Signal
@@ -73,7 +73,7 @@ func init() {
 	flaggy.Bool(&enableForceMaster, "", "forceMaster", "Set to true to enable local testing, forced master mode.")
 	flaggy.Bool(&enableDebug, "d", "debug", "Set to true to enable debug.")
 	flaggy.String(&podCheckNamespaces, "", "podCheckNamespaces", "The comma separated list of namespaces on which to check for pod status and restarts, if enabled.")
-	flaggy.String(&dnsEndpoints, "", "dnsEndpoints", "The comma separated list of dns endpoints to check, if enabled. Defaults to kubernetes.default,cloud.google.com,aws.amazon.com")
+	flaggy.StringSlice(&dnsEndpoints, "", "dnsEndpoints", "The comma separated list of dns endpoints to check, if enabled. Defaults to kubernetes.default")
 	flaggy.Parse()
 
 	// log to stdout and set the level to info by default
@@ -119,9 +119,6 @@ func main() {
 	// Split the podCheckNamespaces into a []string
 	namespaces := strings.Split(podCheckNamespaces, ",")
 
-	// Split the dnsEndpoints into a []string
-	dns := strings.Split(dnsEndpoints, ",")
-
 	// Add enabled checks into Kuberhealthy
 
 	// componentstatus checking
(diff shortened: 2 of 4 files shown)


cmd/kuberhealthy/main.go
```diff
@@ -113,6 +119,9 @@ func main() {
 	// Split the podCheckNamespaces into a []string
 	namespaces := strings.Split(podCheckNamespaces, ",")
 
+	// Split the dnsEndpoints into a []string
+	dns := strings.Split(dnsEndpoints, ",")
```
integrii: Remove this

cmd/kuberhealthy/main.go
```diff
@@ -65,9 +69,11 @@ func init() {
 	flaggy.Bool(&enableDaemonSetChecks, "", "daemonsetChecks", "Set to false to disable cluster daemonset deployment and termination checking.")
 	flaggy.Bool(&enablePodRestartChecks, "", "podRestartChecks", "Set to false to disable pod restart checking.")
 	flaggy.Bool(&enablePodStatusChecks, "", "podStatusChecks", "Set to false to disable pod lifecycle phase checking.")
+	flaggy.Bool(&enableDnsStatusChecks, "", "dnsStatusChecks", "Set to false to disable DNS checks.")
 	flaggy.Bool(&enableForceMaster, "", "forceMaster", "Set to true to enable local testing, forced master mode.")
 	flaggy.Bool(&enableDebug, "d", "debug", "Set to true to enable debug.")
 	flaggy.String(&podCheckNamespaces, "", "podCheckNamespaces", "The comma separated list of namespaces on which to check for pod status and restarts, if enabled.")
+	flaggy.String(&dnsEndpoints, "", "dnsEndpoints", "The comma separated list of dns endpoints to check, if enabled. Defaults to kubernetes.default,cloud.google.com,aws.amazon.com")
```
integrii: recommend using `flaggy.StringSlice()`

integrii: this will do the string split for you, and you can use the flag multiple times with one param each time if you like

pkg/checks/dnsStatus/dnsStatus.go
```diff
@@ -0,0 +1,116 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+// Checker validates that DNS is functioning correctly
+type Checker struct {
+	FailureTimeStamp map[string]time.Time
+	Errors           []string
+	MaxTimeInFailure float64 // TODO - make configurable
+	client           *kubernetes.Clientset
+	Endpoints        []string // TODO - Make configurable
```
integrii: this is no longer `todo`

pkg/checks/dnsStatus/dnsStatus.go
```diff
@@ -0,0 +1,116 @@
...
+	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
+
+	"k8s.io/client-go/kubernetes"
+)
+
+// Checker validates that DNS is functioning correctly
+type Checker struct {
+	FailureTimeStamp map[string]time.Time
+	Errors           []string
+	MaxTimeInFailure float64 // TODO - make configurable
```
integrii: what does this mean? add comment?

pkg/checks/dnsStatus/dnsStatus.go
```diff
@@ -0,0 +1,116 @@
...
+	client           *kubernetes.Clientset
+	Endpoints        []string // TODO - Make configurable
+}
+
+// New returns a new Checker
+func New(endpoints []string) *Checker {
+	defaultEndpoints := []string{
```
integrii: While I think these are good places to resolve, we don't want everyone who uses kuberhealthy to get paged when aws or google goes offline (it should never happen though, right?!)

Perhaps use some root name server hostnames: https://www.iana.org/domains/root/servers ?

We could also make this do internal lookups by default only.  I am sure plenty of clusters don't actually resolve public DNS, or only are allowed to resolve a whitelist of external URLs... 

If we only resolve the internal dns, `kubernetes.default`, that should cover kubernetes.  If something is wrong upstream, then that kind of feels like its out of scope for what we monitor - a different DNS checker for external URLs could be run in the cluster.  I mean to say - we should probably _only_ guarantee that the kubernetes functionality passes.  We probably don't want to consider external resolution at the same time by default...

cmd/kuberhealthy/main.go
```diff
@@ -34,6 +36,7 @@ import (
 var kubeConfigFile = filepath.Join(os.Getenv("HOME"), ".kube", "config")
 var listenAddress = ":8080"
 var podCheckNamespaces = "kube-system"
+var dnsEndpoints = []string{}
```
integrii: you don't need to make a change here, but you could instead write this as `var dnsEndpoints []string`

README.md
```diff
@@ -149,6 +149,18 @@ A command line flag exists `--podCheckNamespaces` which can optionally contain a
...
 
+#### DNS
+
+Checks for failures with DNS, including resolving within the cluster and outside of the cluster. Default endpoints to resolve: kubernetes.default, aws.amazon.com, cloud.dns.com
+
+A command-line flag exists `--dnsEndpoints` which can optionally include a comma separated list of DNS endpoints to test. 
+
+- Timeout: 1 minutes
+- Check Interval: 2 minutes
+- Error state toleration: 5 minutes
```
integrii: I think you removed this five minute state toleration, correct?

ihoegen: Updated it to be a 1 minute toleration

integrii: Why have any time toleration?  I would assume that any dns failure against kubernetes.default would be an error.  Maybe give it a second query on failure to tolerate small hiccups, but failures should be rare and are probably bad.

What do you think?

ihoegen:
I'm not sure if DNS is self healing, that's the main reason behind the
check. Although, if DNS round robins, and there is 1 healthy DNS pod and 1
unhealthy, Kuberhealty would report the state as healthy, even though that
isn't the case. Will remove.

README.md
```diff
@@ -149,6 +149,18 @@ A command line flag exists `--podCheckNamespaces` which can optionally contain a
 - Error state toleration: 5 minutes
 - Check name: `podStatus`
 
+#### DNS
+
+Checks for failures with DNS, including resolving within the cluster and outside of the cluster. Default endpoints to resolve: kubernetes.default, aws.amazon.com, cloud.dns.com
+
+A command-line flag exists `--dnsEndpoints` which can optionally include a comma separated list of DNS endpoints to test. 
+
+- Timeout: 1 minutes
```
lolimjake: If we are going to have such a short error toleration I'd recommend a faster interval.  4-5 checks per minute should be about right.  This is a really light weight check so I don't thing its a problem to check often.

README.md
lolimjake: I mentioned this below, but I agree.  I'd like to run a re-try or a faster interval to verify that the DNS is consistently down before alerting on the state.  Optionally, we could do something like the podRestartChecker and compare DNS responses over time and only alert once some threshold of failures is breached.

integrii:
BTW lets *NOT* merge this into unstable until we cut a release.  I don't want to introduce any instabilities prior to a release.  We've done a ton of testing with the feature set in unstable right now.  Let's milestone this to the next beta release.

ihoegen:
Updated it to run every 15 seconds, and if we have 1 minute of full failures, so 4 failures, it alerts

integrii:
Hold the merge until I cut the release branch next week please. Nice feature!

ihoegen:
We _should_ be good to merge this now right?

PR #90, opened 2019-01-09
Prc statefulset fix

lolimjake:


pkg/checks/podRestarts/podRestarts.go
```diff
@@ -159,46 +154,59 @@ func (prc *Checker) doChecks() error {
...
+		// If the most recent observation restart count is less than the previous observation restart count, delete all the
+		// entries that arent the most recent one, assuming that the pod was restarted but still has the same name.  This is a new
+		// pod with a reset restart counter but has the same exact name as the previous pod that existed but was deleted
+		// we see this in stateful sets
+		if mostRecentObservation.Count < secondMostRecentObservation.Count {
+			// delete all the observations
+			fmt.Println(podName, "was reset to the most recent observation because it has been recreated.")
+			delete(prc.RestartObservations, podName)
+			//Add back in the most recent observation
+			prc.RestartObservations[podName] = append(prc.RestartObservations[podName], mostRecentObservation)
```
integrii: because you just deleted this key in the map, you can just set the results, without an append

pkg/checks/podRestarts/podRestarts_test.go
```diff
@@ -1,18 +1,118 @@
...
+	//TODO check that each pod group has the expected values after reaping and that the mostRecent and second most Recent
+	//TODO values are correctly filled out
+	t.Log("After")
+	for podName, restartObservations := range c.RestartObservations {
+		t.Log(podName)
+		for _, observation := range restartObservations{
+			t.Log(observation)
+		}
+	}
+}
```
integrii: the test never fails with `t.Fail` and thus won't actually be useful in automated testing.  Throw in a couple `if x != y { t.FailNow("test failed")}` style entries to make this really nice

pkg/checks/podRestarts/podRestarts.go
integrii: you could actually not even do a delete, just overwrite with the single entry

lolimjake: I'm deleting the map entry entirely which is a `map[string][]RestartCountObservation` and adding back in a single, non sliced `RestartCountObservation`.  

I could just clear the existing slice (rather than deleting the entire map entry) and append the new `RestartCountObservation`, but to simply overwrite I would have to change the non sliced `RestartCountObservation` into a slice for the storage of only one expected value above.

Either way I'm doing an append and it felt cleaner to just delete the entire map key and replace it since this is a brand new pod.

pkg/checks/podRestarts/podRestarts_test.go
lolimjake: I had a TODO just a few lines above for exactly that.  I'll just put it in this PR though.

Issue #121, opened 2019-02-24
Make dependent images configurable


davidkarlsen:
**Describe the feature you would like and why you want it**
For closed corporate environments any images requested might need to go through an internal proxy, hence it would be good to be able configure the imagerepo/image required by kuberhealthy (like the gcr pause one).
The chart at https://github.com/helm/charts/tree/master/stable/kuberhealthy should take parameters to be able to point at these.

**Additional context**
See https://github.com/Comcast/kuberhealthy/issues/114 for background.

lolimjake:
I've pushed an image tag for you to test with to  ds-image-override.  

I've added this feature as an argument `dsPauseContainerImageOverride` - If you add this argument to your running kuberhealthy deployment at `deployment.spec.template.spec.containers.args` you should be able to specify the location of the pause container image that the daemon set checker deploys.

Ex:

```
spec:
  template:
    spec:
      containers:
        args:
        - "-dsPauseContainerImageOverride"
        - "your-docker-repo/pause:0.8.0"
```

davidkarlsen:
Which image tag should I use?

Small typo, should be
`"your-docker-repo/google_containers/pause:0.8.0"`

davidkarlsen:
doh -  found it - ds-image-override

davidkarlsen:
Sorry to disappoint:
```
kubectl -n kuberhealthy logs kuberhealthy-7fdc478568-pq7v7 
time="2019-02-25T23:02:04Z" level=info msg="Startup Arguments: [/app/kuberhealthy]"
time="2019-02-25T23:02:04Z" level=info msg="Starting web services on port :8080"
time="2019-02-25T23:07:24Z" level=info msg="I am now master. Starting checks."
time="2019-02-25T23:07:24Z" level=info msg="Became master. Starting checks."
time="2019-02-25T23:07:24Z" level=info msg="Running check: PodStatusChecker namespace kube-system"
time="2019-02-25T23:07:24Z" level=info msg="Running check: ComponentStatusChecker"
time="2019-02-25T23:07:24Z" level=info msg="Running check: DaemonSetChecker"
time="2019-02-25T23:07:24Z" level=info msg="Running check: PodRestartChecker namespace kube-system"
time="2019-02-25T23:07:24Z" level=info msg="Searching for unique taints on the cluster."
time="2019-02-25T23:07:24Z" level=info msg="Found taints to tolerate: [{dedicated  infra NoSchedule <nil>}]"
[...]

Issue #114, opened 2019-02-19
panic: runtime error: invalid memory address or nil pointer dereference


davidkarlsen:
**Describe the bug**
POD fails with:
```
kubectl -n kuberhealthy logs kuberhealthy-7868b9cd7b-6s42h 
time="2019-02-19T18:09:48Z" level=info msg="Startup Arguments: [/app/kuberhealthy]"
time="2019-02-19T18:09:48Z" level=info msg="Starting web services on port :8080"
time="2019-02-19T18:09:48Z" level=info msg="I am now master. Starting checks."
time="2019-02-19T18:09:48Z" level=info msg="Became master. Starting checks."
time="2019-02-19T18:09:48Z" level=info msg="Running check: PodStatusChecker namespace kube-system"
time="2019-02-19T18:09:48Z" level=info msg="Running check: DaemonSetChecker"
time="2019-02-19T18:09:48Z" level=info msg="Running check: PodRestartChecker namespace kube-system"
time="2019-02-19T18:09:48Z" level=info msg="Running check: ComponentStatusChecker"
time="2019-02-19T18:09:48Z" level=info msg="Setting state of check ComponentStatusChecker to true []"
time="2019-02-19T18:09:48Z" level=info msg="Searching for unique taints on the cluster."
time="2019-02-19T18:09:48Z" level=info msg="Found taints to tolerate: [{dedicated  infra NoSchedule <nil>}]"
time="2019-02-19T18:09:48Z" level=info msg="Generating daemon set kubernetes spec."
time="2019-02-19T18:09:48Z" level=info msg="Deploying daemon set with tolerations:  [{dedicated  infra NoSchedule <nil>}]"
time="2019-02-19T18:09:48Z" level=info msg="DaemonSetChecker removing 0 daemonset pods"
time="2019-02-19T18:09:48Z" level=info msg="DaemonSetChecker removing daemonset"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xf77cb1]

goroutine 39 [running]:
main.(*Kuberhealthy).setCheckExecutionError(0xc00025e6e0, 0x11b0b5a, 0x10, 0x0, 0x0)
        /go/src/github.com/Comcast/kuberhealthy/cmd/kuberhealthy/kuberhealthy.go:63 +0x211
[...]

PR #124, opened 2019-02-27
Ds image override

lolimjake:
- Allows the user to override the daemon set checker pause container image location - useful if the user is in an environment that forces an image proxy or otherwise wants to use a different image.
- Sets the daemon set checker DS containers to run as user 1000 instead of root - useful for environments that lock down container root users, and just good practice in general.
- Basic test for the above flag
- Documentation for the above flag (and all the rest of kuberhealthy's flags)

See #121 and #114

### DIFF
--- a/docs/FLAGS.md
+++ b/docs/FLAGS.md
@@ -1,0 +1,16 @@
+Available flags for use in Kuberhealthy
+
+# Flags
+
+|Flag|Description|Optional|Default|
+|---|---|---|---|
+|`-kubecfg`|Absolute path to a kube config file.|Yes| `$HOME/.kube/config`|
+|`-listenAddress`|The port kuberhealthy will listen on.|Yes| `8080`|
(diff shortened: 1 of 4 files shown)


pkg/checks/daemonSet/daemonSet_test.go
```diff
@@ -36,6 +36,27 @@ func TestCleanupOrphans(t *testing.T) {
...
+	}
+	if dsc.PauseContainerImage != "gcr.io/google_containers/pause:0.8.0" {
+		t.Fatal("Default Pause Container Image is not set or an unexpected value, actual value:", dsc.PauseContainerImage)
+	}
+
+	dscO, err := New()
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Silly, yes, but this mimics how the program is setting the override value
```
integrii: not sillly!  We are testing the current functionality.  The future may break any unknown functionality that may change behavior.

docs/FLAGS.md
```diff
@@ -0,0 +1,16 @@
+Available flags for use in Kuberhealthy
+
+# Flags
+
+|Flag|Description|Optional|Default|
```
integrii: nice table!

Issue #68, opened 2018-10-16
Add hard pod anti-affinity to specs


integrii:
We should avoid having multiple Kuberhealthy instances on a single node.  This makes the system prone to failure when a single node is lost.  Pod anti-affinity should take care of this.

Brialius:
@integrii hi, is this request for helm chart?

integrii:
Yeah, and the `deploy/kuberhealthy.yaml` file, which should have the same functionality, but for people who don't want to use helm.

PR #131, opened 2019-03-12
[Issue-68] Add hard pod anti-affinity to spec

joshulyne:
Based off of what I read here: 
https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#more-practical-use-cases 
For this issue: 
https://github.com/Comcast/kuberhealthy/issues/68

I think I just specified the podAntiAffinity so that there aren't replicas of Kuberhealthy on a single node -- let me know if this is what you guys were thinking of implementing for this issue and if the implementation is correct!

### DIFF
--- a/deploy/kuberhealthy.yaml
+++ b/deploy/kuberhealthy.yaml
@@ -191,6 +191,16 @@ spec:
     spec:
       serviceAccountName: kuberhealthy
       automountServiceAccountToken: true
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - kuberhealthy
+            topologyKey: "kubernetes.io/hostname"
       containers:
       - image: quay.io/comcast/kuberhealthy:v1.0.2
         imagePullPolicy: IfNotPresent


deploy/kuberhealthy.yaml
```diff
@@ -191,6 +191,16 @@ spec:
     spec:
       serviceAccountName: kuberhealthy
       automountServiceAccountToken: true
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
lolimjake: Since this spec will be installed on unknown clusters, I'd recommend we use `preferredDuringSchedulingIgnoredDuringExecution` with a high weight instead.  Using `requiredDuringSchedulingIgnoredDuringExecution` would prevent Kuberhealthy from spinning up at all on clusters with a single node or clusters that only have a single node with schedulable space which I don't think we want as default behavior. 

You can read about weights [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/nodeaffinity.md#nodeaffinity) but higher = more preference.  An example of the format is [here](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#an-example-of-a-pod-that-uses-pod-affinity).

Other than that, this looks great!  Once the `required`>`preferred` change is made I'll merge it in.

Additionally, we maintain [helm charts](https://github.com/helm/charts/tree/master/stable/kuberhealthy/templates) for Kuberhealthy, you should make a PR for these changes in all the configs there, too!

joshulyne: Yeahhh `preferredDuringSchedulingIgnoredDuringExecution` makes more sense! 

Will create another PR for the changes in the helm chart! Thanks for the speedy review!

joshulyne:
this can also be closed as per #135

Issue #7, opened 2018-06-19
Improve shutdown grace period handling


lolimjake:
Kuberhealthy shutdown grace period should be calculated from the running spec instead of being statically configured.

lolimjake:
#134 Exposes the grace period as an env var, we now just need to consume this var and pass it through to the shutdown contexts

integrii:
Line 39 in `main.go` needs overridden when the environment variable for termination grace period seconds is set.  This probably needs done in the `init()` func.

lolimjake:
https://github.com/Comcast/kuberhealthy/blob/master/cmd/kuberhealthy/main.go#L39

integrii:
Looks like we can't actually do this due to the kubernetes downward API not exposing `terminationGracePeriodSeconds`

PR #132, opened 2019-03-13
[Issue-7] Expose gracePeriodTimeoutSeconds pod spec as env variable

joshulyne:
Addresses part of this issue: https://github.com/Comcast/kuberhealthy/issues/7
- Exposes the pod spec variable`terminationGracePeriodSeconds` as an environment variable `GRACE_PERIOD_TIMEOUT_SECONDS` to the container

Missing the other part of having this value be used as the shutdown grace period instead of the statically configured value.


joshulyne:
i think this pr can be closed now? just made #134 instead!

lolimjake:
Closing, see #134 instead

PR #134, opened 2019-03-13
[Issue-7] Expose gracePeriodTimeoutSeconds pod spec as env variable 

joshulyne:
Addresses part of this issue: #7 

Exposes the pod spec variableterminationGracePeriodSeconds as an environment variable GRACE_PERIOD_TIMEOUT_SECONDS to the container

Missing the other part of having this value be used as the shutdown grace period instead of the statically configured value.


joshulyne:
PR in helm: https://github.com/helm/charts/pull/12192/files

PR #141, opened 2019-03-17
Add InfluxDB metric forwarding

ihoegen:
Opening for code and design review, pending a local test


ihoegen:
Testing complete on my side. To test:

Terminal 1
```
brew install influxdb
influxd
```

Terminal 2
```
influxdb
> create database kuberhealthy
```

Terminal 3
```
cd $GOPATH/src/github.com/Comcast/kuberhealthy/cmd/kuberhealthy
make run-influx
```

Terminal 2
```
> use table kuberhealthy
> show measurements
> select * from DnsStatusChecker_status
```

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -259,6 +261,24 @@ func (k *Kuberhealthy) runCheck(stopChan chan bool, c KuberhealthyCheck) {
 		details := health.NewCheckDetails()
 		details.Namespace = c.CheckNamespace()
 		details.OK, details.Errors = c.CurrentStatus()
+
+		if k.MetricForwarder != nil {
+			checkStatus := 0
```
integrii: var checkStatus also works here and is more normal. no need to change this, though

pkg/metrics/influxdb.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, errors.Wrap(err, "influx.NewClient")
+	}
+	return &InfluxClient{
+		db:     input.Database,
+		client: client,
+	}, nil
+}
+
+// Push pushes a metric of name string and value of any type
+func (i *InfluxClient) Push(name string, value interface{}, tags map[string]string) error {
```
integrii: Maybe refactor to actually take a slice of things in?  You could make one function for a map, then a wrapper function for a single item that creates a map and passes it into the batch, or something like that...

I just feel bad that we're making an `influx.BatchPoints`, but only sending one metric at a time.

PR #146, opened 2019-03-26
influxdb and go module work

integrii:
I need to merge into master (without cutting a release) so that go modules can start working.  Right now, I can't use the modules that are in progress because they don't exist in upstream version control yet.

This may take several revisions to get an initial working module build with all dependencies accounted for.  As you probably know, kuberhealthy won't build right now as-is.


integrii:
I am going to re-open with a cherry-pick branch so we don't merge too much into master.

Issue #163, opened 2019-04-30
Bad/unused environment variables in default deploy specs


ihoegen:
**Describe the bug**
The deploy specs in the repo, *not* in the helm chart, contain these environment variables:
```
          - name: COMPONENT_STATUS_CHECK
            value: true
          - name: DAEMON_SET_CHECK
            value: true
          - name: DNS_STATUS_CHECK
            value: true
          - name: POD_RESTARTS_CHECK
            value: true
          - name: POD_STATUS_CHECK
            value: true
```
which aren't used anywhere in the code, and are invalid because the value must be a string. This causes installation to fail on the deployment. 
**Steps To Reproduce**
`kubectl apply -f deploy/kuberhealthy-and-prometheus.yaml`
**Expected behavior**
Kuberhealthy to install with no errors

integrii:
https://github.com/Comcast/kuberhealthy/pull/164

integrii:
These are now fixed up and/or working.

PR #164, opened 2019-05-01
Fix strings in static yamls

integrii:
Fixing #163


ihoegen-isp:
I don't think those environment variables are even being used, they're supposed to be flags. Can you verify?

integrii:
If they aren't, they will be.  I think @joshulyne was working on adding check on/off flags

ihoegen:
Were check on/off going to be used as flags or environment variables?

integrii:
Both.  Fixed in #165 
Precedence is: flag, env var, default

integrii:
@ihoegen PR plz :-D

Issue #111, opened 2019-02-17
Create helm chart for Kuberhealthy 2


integrii:
The current helm chart enables all checks.  It is possible to pass in flags that disable checks.  For example, if you only wanted the daemonset checker to run, you could pass in helm chart flags that disable the other checks.

This issue has a few items to complete:

- [ ] add input flags to helm chart (in github.com/helm/charts/stable/kuberhealthy)
- [ ] document new helm chart flags on the helm chart readme
- [ ] document the flags used to disable each check in the README.md file where each check is defined

joshulyne:
@integrii @lolimjake created this pr in the helm/charts repo: https://github.com/helm/charts/pull/12863

TBH, took me a bit to understand the implementation part to this task, but is ^^ that what you guys are looking for and is that the way to add input flags to the helm chart? Not sure if there's more I've got to do. 

Lastly, how does setting those values to false get translated to kuberhealthy? Was looking at the kuberhealthy codebase and was trying to figure out where the enable/disable flag would get passed in... let me know if im interpreting this correctly or not! Thanks!!!

integrii:
I left you some comments on your helm PR.

integrii:
This should be held until we refactor the chart for external checks in version 2.  Adding checks will move from flags to specific `khcheck` custom resources.

integrii:
Refactoring this issue for kuberhealthy 2 usefulness

jonnydawg:
Just pushed changes that address this! 
https://github.com/Comcast/kuberhealthy/commit/2bf6631ca33f7febff610b65b2754d3a21e1d185

integrii:
chart PR submitted https://github.com/helm/charts/pull/18899

integrii:
this is being superseded by #288

PR #165, opened 2019-05-01
Check env vars

integrii:
This allows for check enable/disable settings with environment variables.


joshulyne:
thanks @integrii !!! this should finally close out #111 :)

integrii:
> So if I'm understanding correctly, flags override environment variables?

yes

PR #190, opened 2019-09-17
update flaggy that supports csv stringslice

isaaguilar:
Updated the flaggy library with the correct version that support csv `stringSlices`


integrii:
Thanks!!!

PR #206, opened 2019-10-01
merge joshs external ds checker into 2.0.0

integrii:


deploy/kuberhealthy.yaml
```diff
@@ -211,13 +215,13 @@ spec:
...
+            value: "false"
           - name: DNS_STATUS_CHECK
-            value: "true"
+            value: "false"
           - name: POD_RESTARTS_CHECK
-            value: "true"
+            value: "false"
           - name: POD_STATUS_CHECK
-            value: "true"
+            value: "false"
```
integrii: we should set these back to `true` and only keep them `false` in our testing environment.

cmd/daemonSetExternalCheck/daemonSetExternalCheck.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: comcast.github.io/v1
+kind: KuberhealthyCheck
+metadata:
+  name: kh-daemonset-external-check
+  namespace: kuberhealthy
+spec:
+  runInterval: 30s
+  timeout: 15m
```
integrii: i added a timeout here

PR #210, opened 2019-10-04
Catch the PodStatus.Message 

### LINKED ISSUE #201: Detect when external checker images have had an error pulling and report it on the status page
It will be common for users to input a khcheck image with the wrong URL - we should catch this situation of `ErrImagePull` and show it on the status page properly.

u5surf:
Fixes #201 
I implemented that It can be detecting the fail status, and receiving PodStatus.Message as an error.
cf. https://godoc.org/k8s.io/api/core/v1#PodStatus

### DIFF
--- a/pkg/checks/external/main.go
+++ b/pkg/checks/external/main.go
@@ -608,7 +608,14 @@ func (ext *Checker) waitForPodRunning() chan error {
 				}
 
 				// TODO - catch when the pod has an error image pull and return it as an error #201
-
+				for _, containerStat := range p.Status.ContainerStatuses {
+					if containerStat.State.Waiting.Reason == "ErrImagePull" {
+						ext.log("pod had an error image pull")
+						outChan <- errors.New(containerStat.State.Waiting.Reason)
+						watcher.Stop()
+						return
+					}
+				}
 				// read the status of this pod (its ours)
 				ext.log("pod state is now:", string(p.Status.Phase))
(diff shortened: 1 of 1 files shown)


pkg/checks/external/main.go
```diff
@@ -613,7 +613,7 @@ func (ext *Checker) waitForPodRunning() chan error {
 				ext.log("pod state is now:", string(p.Status.Phase))
 				if p.Status.Phase == apiv1.PodRunning || p.Status.Phase == apiv1.PodFailed || p.Status.Phase == apiv1.PodSucceeded {
 					ext.log("pod is now either running, failed, or succeeded")
-					outChan <- nil
+					outChan <- errors.New(p.Status.Message)
```
integrii: This function (`waitForPodRunning()`) is actually supposed to not return an error once the pod stops running - for any reason.  An error here would mean that we failed to communicate with the Kubernetes API, or had some other internal failure.

integrii:
I think what we want to do here is not to return the error as proposed, but instead find a way to detect when the pod enters a state of `ErrImagePull`.  When that happens, we should craft a specific error, kind of like you did here, and return that.

u5surf:
We might have caught this message to iterate `Pod.Status.ContainerStautses`
e.g.
https://github.com/ayufan/gitlab-ci-multi-runner/blob/a3d5829a46cb94a49c438db7efede4da40a0a238/executors/kubernetes/util.go#L112-L124

PR #217, opened 2019-10-07
fix typo in podStatus checks

isaaguilar:
`podStatus` checker was always purging the `previouslyFailedContainer` because of a typo in `currentlyFailedContainer`


integrii:
Nice catch!  I am going to merge this into the v2.0.0 branch as well so that we can ensue this fix is ported to an external check properly.

PR #218, opened 2019-10-09
Remove package lists after updating in Dockerfile #8

amandahla:
Change to all the apt commands in the same layer and remove package lists/cache. Image size: 858MB.


integrii:
Very helpful! Thanks!

PR #224, opened 2019-10-17
Refactor and externalize pod restarts check

joshulyne:


cmd/podRestartsExternalCheck/main.go
```diff
@@ -0,0 +1,263 @@
...
+	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	checkclient "github.com/Comcast/kuberhealthy/pkg/checks/external/checkClient"
+	"github.com/Comcast/kuberhealthy/pkg/kubeClient"
+)
+
+const maxFailuresAllowed = 5
```
joshulyne: not sure if we want to make this configurable and set it as an optional env var with a default set to 5?

cmd/podRestartsExternalCheck/podRestartsExternalCheck.yaml
```diff
@@ -0,0 +1,22 @@
...
+  runInterval: 60m
+  timeout: 58m
+  podSpec:
+    containers:
+      - env:
+          - name: POD_NAMESPACE
+            value: "kube-system"
+          - name: CHECK_RUN_WINDOW
+            value: "55m"
+        image: docker-proto.repo.theplatform.com/pod-restarts-check:1.0.0
```
joshulyne: need to change image location

Makefile
```diff
@@ -14,3 +14,11 @@ buildExternalChecker:
...
 pushExternalChecker:
 	docker push integrii/test-external-check:latest
+
+podRestarts: buildPodRestartsCheck pushPodRestartsCheck
+
+buildPodRestartsCheck:
+	docker build -t docker-proto.repo.theplatform.com/pod-restarts-check:1.0.0 -f cmd/podRestartsExternalCheck/Dockerfile .
+
+pushPodRestartsCheck:
+	docker push docker-proto.repo.theplatform.com/pod-restarts-check:1.0.0
```
joshulyne: need to build and push somewhere public

cmd/podRestartsExternalCheck/main.go
integrii: i would definitely like to see this configurable.  Here is an issue for it: #225

cmd/podRestartsExternalCheck/podRestartsExternalCheck.yaml
integrii: I have invited you to the quay.io account.  You can make a new public repository there and push these images in.

Makefile
integrii: let's wait for this image to be hosted in our quay.io account before merging so we don't forget

cmd/podRestartsExternalCheck/main.go
```diff
@@ -0,0 +1,297 @@
...
+		length := 0
+		BadPodRestarts.Range(func(_, _ interface{}) bool {
+			length++
+			return true
+		})
+
+		if length > 0 {
+			log.Println("Bad pod with restarts found, reporting check failure to Kuberhealthy servers")
+
+			prc.errorMessages = nil //clear error messages before every failure report
```
integrii: re-init to a blank slice here, instead of a nil `[]string{}`

cmd/podRestartsExternalCheck/main.go
```diff
@@ -0,0 +1,297 @@
...
+
+			BadPodRestarts.Range(func(k, v interface{}) bool {
+
+				errorMessage := "Pod restarts for pod: " + k.(string) + " is greater than " + strconv.Itoa(int(prc.MaxFailuresAllowed)) + " in the last hour. Restart Count is at: " + strconv.Itoa(int(v.(int32)))
+				prc.errorMessages = append(prc.errorMessages, errorMessage)
+
+				return true
+			})
+			err = checkclient.ReportFailure(prc.errorMessages)
```
integrii: make sure the `len(prc.errorMessages)` is greater than 0

PR #226, opened 2019-10-19
add deployment check

### LINKED ISSUE #154: Implement deployment checker
**Describe the solution you would like to see happen**
A check for Deployments analogous to the existing DaemonSet check would be useful to verify, that the Controller Manager loops for Deployments and ReplicaSet are still working. As most applications utilize Deployments, it is important to be sure, that this functionality is working.

**How do you propose we build it?**
The check could be built similar to the existing check for daemon set, but using a Deployment instead of a DaemonSet. One replica would be enough for this check, as the DaemonSet check already verifies, that pods can be scheduled on all nodes.

What is your opinion on deployment checks? Would you accept contributions for such a check?

jonnydawg:
Adds external deployment check! 
Fixes #154


README.md
```diff
@@ -150,6 +150,16 @@ A command-line flag exists `--dnsEndpoints` which can optionally include a comma
 - Check name: `dnsStatus`
 
 
+#### Deployment and Service
+
+Deploys a `deployment` to the `kuberhealthy` namespace with `2` replicas, waits for pods and service to be in the 'Ready' state and serving a hostname, then makes a request to the endpoint for a `200 OK`, then terminates them and ensures deployment and service terminations were successful.  Containers are deployed with their resource requirements set to 15 millicores and 20Mi units of memory.  The check uses the latest Nginx web server container (`nginx:latest`) for the deployment.  A success means that a deployment and service can be brought up and that the corresponding hostname endpoint can return a `200 OK`.  If `CHECK_DEPLOYMENT_ROLLING_UPDATE` is set to `true`, the check will initially deploy Nginx's alpine web server container and roll to the latest (`nginx:alpine` -> `nginx:latest`) and look for a second `200 OK`.  A failure means that an error occurs anywhere in the deployment creation, service creation, HTTP request, and tear down resulting in an error report to Kuberhealthy.
```
joshulyne: --> The check deploys a `deployment` to the `kuberhealthy` namespace with `2` replicas and waits for both the pods and service to be in `Ready` state and serve a hostname. Once ready, the check makes a request to the endpoint for a `200 OK` response and then proceeds to terminate them and ensures that the deployment and service terminations were successful. 

--> A successful check run implies that a deployment and service can be brought up and the corresponding hostname endpoint returns a `200 OK` response. 

--> and look for a second `200 OK` response. 

--> A failure implies that an error occurred anywhere in the deployment creation, service creation, HTTP request, or tear down process, resulting in an error report to the Kuberhealthy status page.

cmd/deployment-check/deployment-check.yaml
```diff
@@ -0,0 +1,28 @@
...
+      imagePullPolicy: IfNotPresent
+      env:
+        - name: CHECK_DEPLOYMENT_REPLICAS
+          value: "6"
+        - name: CHECK_TIME_LIMIT_SECONDS
+          value: "300"
+        - name: CHECK_DEPLOYMENT_ROLLING_UPDATE
+          value: "true"
+        - name: SHUTDOWN_GRACE_PERIOD_SECONDS
+          value: "30"
```
joshulyne: maybe explain these env vars either here or in your readme?

cmd/deployment-check/deployment.go
```diff
@@ -0,0 +1,703 @@
...
+	// Default container resource requests values.
+	defaultMillicoreRequest = 15               // Calculated in decimal SI units (50 = 50m cpu).
+	defaultMillicoreLimit   = 75               // Calculted in decimal SI units (100 = 100m cpu).
+	defaultMemoryRequest    = 20 * 1024 * 1024 // Calculated in binary SI units (50 * 1024^2 = 50Mi memory).
+	defaultMemoryLimit      = 60 * 1024 * 1024 // Calculated in binary SI units (100 * 1024^2 = 100Mi memory).
+
+	// Default container probe values.
+	defaultProbeFailureThreshold    = 5  // Number of consecutive failures for the probe to be considered failed (k8s default = 3).
+	defaultProbeSuccessThreshold    = 1  // Number of consecutive successes for the probe to be considered successful after having failed (k8s default = 1).
+	defaultProbeInitialDelaySeconds = 2  // Number of seconds after container has tarted before probes are initiated.
```
joshulyne: tarted --> started

cmd/deployment-check/main.go
```diff
@@ -0,0 +1,259 @@
...
+	ctxCancel() // Causes all functions within the check to return without error and abort. NOT an error
+	// condition; this is a response to an external shutdown signal.
+
+	// Clean up pods here.
+	log.Infoln("Shutting down.")
+
+	select {
+	case <-signalChan:
+		// If there is an interrupt signal, interrupt the run.
+		log.Warnln("Received a secsond interrupt signal from the signal channel.")
```
joshulyne: secsond -> second

cmd/deployment-check/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.13 AS builder
+RUN apt-get upgrade -y && apt-get install git -y
+ADD . /go/src/app
+WORKDIR /go/src/app
+RUN go get -v
+RUN go build -o app
+RUN mkdir /app
+RUN mv app /app/app
+
+FROM golang:1.13
```
integrii: Move to an alpine image here.  You may need to set `GOOS=linux` in the builder and will need to `apk update && apk add ca-certificates`.

cmd/deployment-check/deployment-check.yaml
```diff
@@ -0,0 +1,28 @@
...
+      resources:
+        requests:
```
integrii: consider adding limits here for cpu (maybe not memory to prevent unnecessary crashes)

cmd/deployment-check/deployment-check.yaml
integrii: looks like he did do this

cmd/deployment-check/deployment.go
```diff
@@ -0,0 +1,703 @@
...
+			// Stop the watch on each loop because we will create a new one.
+			watch.Stop()
+		}
+	}()
+
+	return createChan
+}
+
+// createContainerConfig creates a container resource spec and returns it.
+func createContainerConfig(imageURL string) corev1.Container {
```
integrii: this is a good candidate for a unit test

cmd/deployment-check/deployment.go
```diff
@@ -0,0 +1,703 @@
...
+			case <-ctx.Done():
+				log.Errorln("context expired while waiting for service to be cleaned up.")
+				cleanUpChan <- nil
+				return
+			}
+		}
+	}()
+
+	log.Infoln("Removing previous deployment in", checkNamespace, "namespace.")
+	if client == nil {
```
integrii: check this upstream rather than in all downstream usages

cmd/deployment-check/deployment.go
```diff
@@ -0,0 +1,704 @@
...
+			result.Err = err
+			updateChan <- result
+			return
+		}
+
+		// Stop the watch on each loop because we will create a new one.
+		defer watch.Stop()
+
+		log.Debugln("Watching for deployment rolling-update to complete.")
+		newPodStatuses := make(map[string]bool)
```
integrii: consider a refactor that that goes off of the status on a deployment instead of a watch with your own state in a map.  This way the deployment logic is exactly aligned with your determination of deployment success.

image

cmd/deployment-check/deployment.go
```diff
@@ -0,0 +1,704 @@
...
+}
+
+// waitForDeploymentToDelete waits for the service to be deleted.
+func waitForDeploymentToDelete() chan bool {
+
+	// Make and return a channel while we check that the service is gone in the background.
+	deleteChan := make(chan bool, 1)
+
+	go func() {
+		defer close(deleteChan)
```
integrii: sleep in here so not to dos the API

cmd/deployment-check/main.go
```diff
@@ -0,0 +1,259 @@
...
+	case err := <-cleanUpAndWait():
+		// If the clean up is complete, exit.
+		log.Infoln("Received a complete signal, clean up completed.")
+		if err != nil {
+			log.Errorln("failed to clean up check resources properly:", err.Error())
+		}
+	case <-time.After(time.Duration(shutdownGracePeriodSeconds)):
```
integrii: make sure that `shutdownGracePeriodSeconds` always aligns with the `khcheck` pod spec.  You can define this variable in there and it defaults to 30s.

cmd/deployment-check/run_check.go
```diff
@@ -0,0 +1,290 @@
...
+	"errors"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// runDeploymentCheck sets up a deployment and applies it to the cluster.
+// Sets up a deployment and service and uses the client to deploy the test deployment and service.
+// Attempts to hit the service hostname endpoint, looking for a 200 and reports a success if able to retrieve a 200.
+func runDeploymentCheck() error {
```
integrii: there isn't a case where returning an error is valuable in this code, so remove the `error` return from the signature

cmd/deployment-check/run_check.go
```diff
@@ -0,0 +1,290 @@
...
+		return nil
+	}
+
+	// If a rolling update is enabled, perform a rolling update on the service.
+	if rollingUpdate {
+
+		log.Infoln("Rolling update option is enabled.  Performing roll.")
+
+		// Create a rolling-update deployment resource.
+		rolledUpdateConfig := createDeploymentConfig(true)
```
integrii: change `createDeploymentConfig` to not take a bool and instead do the one-time modification of the deployment inline here

cmd/deployment-check/service.go
```diff
@@ -0,0 +1,366 @@
...
+				log.Debugln("Received an event watching for service changes:", event.Type)
+
+				s, ok := event.Object.(*corev1.Service)
+				if !ok {
+					log.Debugln("Got a watch event for a non-service object -- ignoring.")
+					continue
+				}
+
+				// We want an event type of DELETED here.
+				if event.Type == "DELETED" {
```
integrii: consider moving to a constant based on the `watch` package

cmd/deployment-check/service.go
```diff
@@ -0,0 +1,366 @@
...
+// waitForServiceToDelete waits for the service to be deleted.
+func waitForServiceToDelete() chan bool {
+
+	// Make and return a channel while we check that the service is gone in the background.
+	deleteChan := make(chan bool)
+
+	go func() {
+		defer close(deleteChan)
+		for {
+			_, err := client.CoreV1().Services(checkNamespace).Get(checkServiceName, metav1.GetOptions{})
```
integrii: needs a 😴💤

jonnydawg: added here and in deployment.go!

cmd/deployment-check/service.go
jonnydawg: Changes made in both service.go and deployment.go

cmd/deployment-check/run_check.go
jonnydawg: This now takes a string

cmd/deployment-check/run_check.go
jonnydawg: removed

cmd/deployment-check/deployment.go
jonnydawg: removed

cmd/deployment-check/deployment.go
jonnydawg: basic tests for config creation have been added

cmd/deployment-check/deployment-check.yaml
jonnydawg: Added a limit here, but only for cpu

PR #236, opened 2019-10-24
Add tests to daemonset checker

joshulyne:
Also fixed a test added to the pod-restarts check


cmd/daemonSetCheck/Dockerfile
```diff
@@ -1,9 +1,11 @@
-FROM golang:1.13 AS builder
+FROM golang:1.13-alpine AS builder
+RUN apk upgrade
 ADD . /build
 WORKDIR /build/cmd/daemonSetCheck
+CMD CGO_ENABLED=0 go test ./...
```
integrii: always run `go test` with `-v` in dockerfiles so you know what goes wrong in the build log.  Also, why are you using `CGO_ENABLED=0` here?

joshulyne: cause i get this error if i don't enable cgo
```# runtime/cgo
exec: "gcc": executable file not found in $PATH```

File: cmd/daemonSetCheck/Dockerfile [hunk 2]
```diff
@@ -1,9 +1,12 @@
-FROM golang:1.13 AS builder
+FROM golang:1.13-alpine AS builder
+RUN apk upgrade
```
jonnydawg: Is `RUN apk upgrade` necessary here? You're upgrading alpine packages in the builder and then throwing it away. The go binary is all that will be running in this container.

cmd/podRestartsCheck/Dockerfile
```diff
@@ -1,8 +1,11 @@
-FROM golang:1.13 AS builder
+FROM golang:1.13-alpine AS builder
+RUN apk upgrade
```
jonnydawg: Same thing here

cmd/daemonSetCheck/Dockerfile
integrii: it could matter because it will cause the binary to be produced with the newest versions of build software and libraries.  I would leave it in, I think.

cmd/podRestartsCheck/README.md
```diff
@@ -44,5 +44,8 @@ spec:
 
 #### How-to 
 
-To implement the Pod Restarts Check with Kuberhealthy, apply the configuration file [podRestartsCheck.yaml](podRestartsCheck.yaml) to your Kubernetes Cluster.
+To implement the Pod Restarts Check with Kuberhealthy, run:
+ 
+`kubectl apply -f https://raw.githubusercontent.com/Comcast/kuberhealthy/2.0.0/cmd/podRestartsCheck/podRestartsCheck.yaml`
```
integrii: I like having easy-to-paste commands like this in our documentation.

joshulyne:
@integrii yupp, when i run `make build<CheckName>` for all three checks (Daemonset, DNSStatus, PodRestarts), all tests pass

Issue #205, opened 2019-10-01
Improve error handling for duration parsing


integrii:
When a khcheck check has an invalid duration or timeout set, we assume a default and move on.  See line 314 in `kuberhealthy.go`.

We should change this to set a `khstate` error if the fields fail to parse.  The check should _not_ be added to the list of configured checks in this case.

integrii:
https://github.com/Comcast/kuberhealthy/blob/2.0.0/cmd/kuberhealthy/kuberhealthy.go#L314

joshulyne:
https://github.com/Comcast/kuberhealthy/blob/2.0.0/cmd/kuberhealthy/kuberhealthy.go#L344

integrii:
this was fixed at some point in our error handling improvements

PR #239, opened 2019-10-24
Updating error handling for duration parsing for issue 205.

teracross:
For issue #205

### DIFF
--- a/cmd/kuberhealthy/kuberhealthy.go
+++ b/cmd/kuberhealthy/kuberhealthy.go
@@ -326,6 +326,8 @@ func (k *Kuberhealthy) addExternalChecks() error {
 
 	// iterate on each check CRD resource and add it as a check
 	for _, r := range l.Items {
+		var isZero bool = false
+
 		log.Debugln("Loading check CRD:", r.Name)
 
 		log.Debugf("External check custom resource loaded: %v", r)
@@ -334,12 +336,11 @@ func (k *Kuberhealthy) addExternalChecks() error {
 		log.Infoln("Enabling external check:", r.Name)
 		c := external.New(kubernetesClient, &r, khCheckClient, khStateClient, externalCheckReportingURL)
 
-		// parse the run interval string from the custom resource and setup the run interval
+		// parse the run interval string from the custom resource and setup the run interval to use the default zero interval
 		c.RunInterval, err = time.ParseDuration(r.Spec.RunInterval)
 		if err != nil {
 			log.Errorln("Error parsing duration for check", c.CheckName, "in namespace", c.Namespace, err)
 			log.Errorln("Defaulting check to a runtime of ten minutes.")
-			c.RunInterval = DefaultRunInterval
 		}
 
 		log.Debugln("RunInterval for check:", c.CheckName, "set to", c.RunInterval)
@@ -361,7 +362,9 @@ func (k *Kuberhealthy) addExternalChecks() error {
 		c.ExtraLabels = r.Spec.ExtraLabels
 
 		// add the check into the checker
-		k.AddCheck(c)
+		if ( !isZero ) {
+			k.AddCheck(c)
+		}
 	}
 
 	return nil
--- a/cmd/kuberhealthy/kuberhealthy.go
+++ b/cmd/kuberhealthy/kuberhealthy.go
@@ -326,8 +326,6 @@ func (k *Kuberhealthy) addExternalChecks() error {
 
 	// iterate on each check CRD resource and add it as a check
 	for _, r := range l.Items {
-		var isZero bool = false
-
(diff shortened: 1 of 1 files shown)


teracross:
After testing with @joshulyne, the feature seems to work as expected when we have have a working configuration and then re-apply with either an invalid timeout or duration value. However upon further testing we it seems when the following combinations for valid/invalid timeout happen the updates and deletes are not made as expected: 

valid interval/ timeout ->(kubectl apply) -> valid interval/ timeout ->(kubectl apply) -> invalid  interval/ timeout -> khcheck  pods restarted as if configuration was valid

invalid interval/ timeout ->(kubectl apply) -> invalid interval/ timeout ->(kubectl apply) -> valid  interval/ timeout -> khcheck pods did NOT restart as if configuration was invalid

teracross:
As a side note, we also noticed that https://github.com/Comcast/kuberhealthy/blob/2.0.0/pkg/checks/external/main.go#L105 is assigning a default value to the RunTimeout. Is this default value still needed?

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -334,24 +334,22 @@ func (k *Kuberhealthy) addExternalChecks() error {
...
-		// parse the run interval string from the custom resource and setup the run interval
+		// parse the run interval string from the custom resource and setup the run interval, error if interval not set properly
 		c.RunInterval, err = time.ParseDuration(r.Spec.RunInterval)
 		if err != nil {
-			log.Errorln("Error parsing duration for check", c.CheckName, "in namespace", c.Namespace, err)
-			log.Errorln("Defaulting check to a runtime of ten minutes.")
-			c.RunInterval = DefaultRunInterval
+			errorMessage := "Error parsing duration for check " + c.CheckName + " in namespace " + c.Namespace + err.Error()
+			log.Errorln(errorMessage)
+			return errors.New(errorMessage)
```
integrii: If we return here, then we halt adding external checks.  I would prefer that when we have an external check configuration that isn't good (for any reason, including the RunInterval), we just log an error state for that specific check.  This will keep other checks running and let the user know that this one check configuration is invalid (and the check isn't running).

Check out [this code](https://github.com/Comcast/kuberhealthy/blob/2.0.0/cmd/kuberhealthy/kuberhealthy.go#L509) as an example here.  In this codepath, we build a `details` variable that contains the `khstate` of the check as we want it to be set.  Then, we call `k.storeCheckState()` on it to write it into the cluster's `khstate` resources.

I would expect a similar flow here when we fail to parse the configurations for the check.

It should be noted that if we do not continue on to run `k.AddCheck(c)` below, this check will _not_ appear on the status page or in metrics.  It will simply disappear because the `Kuberhealthy` struct isn't configured with it.  On the flip side, if we do run `k.AddCheck(c)`, then the check will end up being fed into the check runner code [here](https://github.com/Comcast/kuberhealthy/blob/2.0.0/cmd/kuberhealthy/kuberhealthy.go#L444).  Notice the first thing it does is setup a ticker with the `c.RunInterval`.

Perhaps we add an extra property to `Kuberhealthy` that contains a `map[string][]string`.  This would contain all the checks that should be skipped in each namespace (by name).  When a check fails to parse for any reason, we add its name to the list.  When running checks, we skip any check name found in the map.  Each time we re-configure Kuberhealthy's checks, we build a new map of checks to skip so not to get the map into a bad state.

This would make it possible to both add the check to the `Kuberhealthy.Checks` slice so it appears on the status page, AND to not attempt to actually execute the check.



So , to recap, here is how I propose this be written:
- Add a new `skippedChecks` `map[string][]string` property to `Kuberhealthy` and be sure it is initialized when checks are being added [here](https://github.com/Comcast/kuberhealthy/blob/2.0.0/cmd/kuberhealthy/kuberhealthy.go#L942) (`make map[string][]string`)
- Add a new method to `Kuberhealthy` that checks the new `skippedChecks` `map[string][]string` for a specific check in a specific namespace (just return a bool)
- Add a new method to `Kuberhealthy` that adds a check to be skipped to the `skippedChecks` namespace when given a namespace string and a check name string.
- Use the new method to decide if a check should be skipped on each run [here](https://github.com/Comcast/kuberhealthy/blob/2.0.0/cmd/kuberhealthy/kuberhealthy.go#L446).  Add a sleep and `continue` so it does not loop too fast.
- When a check [fails to parse something in around here](https://github.com/Comcast/kuberhealthy/blob/2.0.0/cmd/kuberhealthy/kuberhealthy.go#L338) during `configureChecks`, use the method you created to add the check by its namespace and name to the `skippedChecks` map.

integrii:
this was fixed during a bunch of other improvements lately

Issue #249, opened 2019-11-05
KH_POD_NAMESPACE being applied multiple times on pods


integrii:
```
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
time="2019-11-05T00:56:12Z" level=debug msg="Checking environment variable on calling pod: KH_POD_NAMESPACE "
[...]

PR #250, opened 2019-11-05
2.0.0 pod namespace env

jonnydawg:
Addresses #249 

Resets the injected `KH_POD_NAMESPACE` env var.


integrii:
In the future, we may want to consider a refactor of this in such a way that a single func can be called to both purge and set the proper environment variables.

Then, we wouldn't have to maintain a list of vars that we purge - we just loop over all the ones we want to add and the purge is automatic based on the var being enforced upon the pod spec.

PR #266, opened 2019-11-15
2.0.0

integrii:
Kuberhealthy 2.  

- new `khcheck` resource
- new external check system implemented
- all internal checks except "external" removed
- caching of khstate resources used on the status page
- Helm 3
- New readme
- ability to create and run your own checker containers
- greatly improved master election logic
- greatly improved check reconfiguration logic
- new deployment checker

It is not possible to directly update from Kuberhealthy 1.  You must learn about KH2's new external checks and enable the ones you would like to use.  By default, this release only installs the daemonset checker and deployment checker.


deploy/helm/kuberhealthy/templates/service.yaml
```diff
@@ -0,0 +1,26 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: {{ template "kuberhealthy.name" . }}
+  name: {{ template "kuberhealthy.name" . }}
+  {{- if or .Values.prometheus.enabled .Values.service.annotations }}
+  annotations:
+  {{- if .Values.prometheus.enabled }}
```
jonnydawg: I've set up prometheus to scrape the service in front of the pods here. Maybe remove the one from the deploy and keep this?

deploy/helm/kuberhealthy/templates/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+      {{- range $key, $value := .Values.deployment.podAnnotations }}
+        {{ $key }}: {{ $value | quote }}
+      {{- end }}
+      {{- end }}
+      {{- if .Values.prometheus.enabled -}}
+      {{- if .Values.prometheus.enableScraping -}}
+      {{- if not .Values.deployment.podAnnotations }}
+      annotations:
+      {{- end }}
+        prometheus.io/scrape: "true"
```
jonnydawg: Prometheus scrape annotations are also on the service -- this might be unnecessary to scrape pods here.

cmd/test-external-check/README.md
```diff
@@ -0,0 +1,66 @@
...
+      name: main
+      resources:
+        requests:
+          cpu: 10m
+          memory: 50Mi
+```
+
+Apply this check in your cluster (normally for testing) by running `kubectl apply -f khcheck.yaml`.
+
+**Pro tip: by setting the `imagePullPolicy` to `Never` on your check's spec or on the kuberhealthy deoplyment in your test environment, the Docker Kubernetes cluster will pull the image from your local docker host instead of the web.  This enables rapid development when doing local builds.**
```
joshulyne: We should reference this readme/doc somewhere in the main readme so it's easier to find for folks who want to experiment with the new KH2. Don't think we have it referenced anywhere?

joshulyne: I feel like this `cmd/test-external-check` directory could provide good material for folks to test kuberhealthy locally if they want to just mess around with it -- this documentation would have to get updated though

integrii:
> General comment, should we not include / remove checks that aren't ready to be implemented? Such as the podRestarts check or the podStatus check and move them to another branch / pr?
> 
> This should also be reflected in our external checks registry.

These checks are not enabled in the helm chart or flat file specs, so they won't be enabled by defult for users.

cmd/test-external-check/README.md
integrii: There is a link to this check code in the `EXTERNAL_CHECK_CREATION.md` file, which is linked from the main page.

Main readme -> make your own checks help -> example check

PR #276, opened 2019-12-05
HTTP check

jonnydawg:
Simple GET request check.

### DIFF
--- a/cmd/http-check/main.go
+++ b/cmd/http-check/main.go
@@ -46,7 +46,7 @@ func main() {
 		os.Exit(0)
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/http-check/main.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	// Check if the response status code is a 200.
+	if r.StatusCode == http.StatusOK {
+		log.Infoln("Got a", r.StatusCode, "with a", http.MethodGet, "to", checkURL)
+		kh.ReportSuccess()
+		os.Exit(0)
+	}
+
+	reportErr := fmt.Errorf("unable to retrieve a " + strconv.Itoa(http.StatusOK) + " from " + checkURL)
```
joshulyne: not sure if it'd be helpful to log the response code if its not 200 so users know what the response was?

Issue #281, opened 2019-12-10
Add check run duration metric to khstate


joshulyne:
Add another metric to Kuberhealthy in KHStates: check duration -- how long the check took to run (success of failure)

PR #282, opened 2019-12-11
Add run duration to khstate

joshulyne:
Addresses issues #281 
Adds check run duration to KHState.

Not sure about the image tag. Made it 2.1.2 for now.


cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -629,6 +631,12 @@ func (k *Kuberhealthy) runCheck(ctx context.Context, c KuberhealthyCheck) {
 		}
 		log.Debugln("Done running check:", c.Name(), "in namespace", c.CheckNamespace())
 
+		// Record check run end time
+		// Subtract 10 seconds from run time since there are two 5 second sleeps during the check run where kuberhealthy
+		// waits for all pods to clear before running the check and waits for all pods to exit once the check has finished
+		// running. Both occur before and after the checker pod completes its run.
+		checkRunDuration := time.Now().Sub(checkStartTime) - time.Second * 10
```
integrii: This worries me because if the upstream delays change in the future, then this logic will break and make bad times.  I think the right solution is to refactor the things that cause delay so that they don't do polling anymore.  That has not been an easy change in the past, though, so for now I think this will have to do...

Makefile
```diff
@@ -1,5 +1,5 @@
 IMAGE="quay.io/comcast/kuberhealthy"
-TAG="2.0.0"
+TAG="2.1.2"
```
integrii: lets just call this 2.1 because it will go out with the other 2.1 features.

Issue #280, opened 2019-12-10
Add optional filter to json status page


integrii:
It would be great to specify a namespace on the json status page in order to easily filter the output of checks.

This would allow developers to share the json status page gracefully instead of having the global `OK` bool flipped whenever anyone's check in the entire cluster goes down.

PR #283, opened 2019-12-11
Status page

jonnydawg:
Status page now handles namespaces requests and will only show checks from the requested namespace.

 An empty namespace request will show checks from the `kube-system` and `kuberhealthy` namespaces. 

#280


cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1010,8 +1010,12 @@ func (k *Kuberhealthy) healthCheckHandler(w http.ResponseWriter, r *http.Request
 		return err
 	}
 
+	// get URL query parameters if there are any
+	values := r.URL.Query()
+	namespace := values.Get("namespace")
```
joshulyne: what happens is there isn't any "namespace" value in the query?

joshulyne: ah just checked the `values.Get` command, it does return an empty string

joshulyne: looks good! 👍

joshulyne:
lgtm!

jonnydawg:
Where should these changes be documented?

cmd/kuberhealthy/reflector.go
```diff
@@ -51,8 +51,8 @@ func (sr *StateReflector) Start() {
 	sr.reflector.Run(sr.reflectorSigChan)
 }
 
-// CurrentStatuses returns the current summary of all checks as known by the cache
-func (sr *StateReflector) CurrentStatus() health.State {
+// CurrentStatus returns the current summary of all checks as known by the cache
```
integrii: What happens if I want to pull the state for all checks?  Do I use a blank input parameter?  How do I know to do that?

Add some details to the comment here about how you can use this to poll everything.

cmd/kuberhealthy/reflector.go
```diff
@@ -74,6 +74,22 @@ func (sr *StateReflector) CurrentStatus() health.State {
...
 
+		// if there is a requested namespace, then filter out checks from other namespaces
+		if len(namespace) != 0 {
+			if khState.GetNamespace() != namespace {
+				log.Debugln("Skipping", khState.GetName(), "because it is not from the", namespace, "namespace")
+				continue
+			}
+		}
+
+		// if there is no requested namespace, show checks from "kube-system" and "kuberhealthy"
```
integrii: this is a preference to us as the project owners, but is an unlikely assumption for all other users.  I would make the default "blank" functionality pick everything.

PR #284, opened 2019-12-13
Exit on failure

jonnydawg:
Sometimes this check runs into a nil pointer crash, this should address that.
Add an exit path on failures.


cmd/http-check/main.go
```diff
@@ -37,6 +37,7 @@ func main() {
 		reportErr := fmt.Errorf("error occurred performing a " + http.MethodGet + " to " + checkURL + ": " + err.Error())
 		log.Errorln(reportErr.Error())
 		kh.ReportFailure([]string{reportErr.Error()})
+		os.Exit(0)
```
integrii: I would take the return from `kh.ReportFailure` and exit `1` or `0` based on if an error was returned.

```
if err != nil {
    os.Exit(1)
}
os.Exit(0)

integrii: you're just exiting 0 here every time - check my code above ^^^

PR #289, opened 2020-01-09
Deployment check patch

jonnydawg:
Updated the way the check handles detection during the clean up process. Moved from a `for loop recreating watches` to a single `watch`.


jonnydawg:
This change removes the watch completely and replaces it with polling lists and checking if the service / deployment exists within the list.

PR #290, opened 2020-01-13
Use non-deprecated apis for daemonset

stenwt:
The extensions/v1beta1 endpoint referenced here is deprecated and removed in K8s 1.16: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

The new endpoint is apps/v1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,3 +70,11 @@ buildPodStatusCheck:
 
 pushPodStatusCheck:
 	docker push quay.io/comcast/pod-status-check:1.0.1
+
+httpCheck: buildHTTPCheck pushHTTPCheck
+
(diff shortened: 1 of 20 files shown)


integrii:
Thanks again @stenwt !  If you want to use these changes, along with the ones coming, please use the image `quay.io/comcast/kuberhealthy:2.1.0rc1`.  Take note that there is now a new spec that needs applied (the check reaper) which is in the 2.1 branch here: https://github.com/Comcast/kuberhealthy/blob/2.1/deploy/kuberhealthy.yaml#L400

You'll need all 3 of the last specs listed there referencing `check-reaper`

PR #304, opened 2020-02-13
fix DSPauseContainerImageOverride

qqshfox:
`pause` image can't be overridden since `DSPauseContainerImageOverride` is not used.


integrii:
For the record, I think this just got dropped when the check was moved to an external khcheck.

PR #305, opened 2020-02-13
Fix dscheck permission

### LINKED ISSUE #300: DaemonSetChecker is failing
**Describe the bug**
DaemonSetChecker is failing because of insufficient permission for the user

**Steps To Reproduce**
Apply the YAML from v2.1.0
Check DaemonSetChecker pod logs 

**Expected behavior**
insufficient permission for the user

**Screenshots**

Screenshot 2020-02-06 at 12 15 31

Screenshot 2020-02-06 at 12 15 53

**Versions**
 - Cluster OS: Ubuntu 18.04
 - Kubernetes Version: 1.11
 - Kuberhealthy Release or build [v2.1.0]

qqshfox:
fix #300

### DIFF
--- a/deploy/kuberhealthy-prometheus-operator.yaml
+++ b/deploy/kuberhealthy-prometheus-operator.yaml
@@ -59,6 +59,7 @@ metadata:
     prometheus: prometheus
     role: alert-rules
   name: kuberhealthy
+  namespace: kuberhealthy
 data:
   kuberhealthy.rules: |-
     groups:
@@ -486,6 +487,7 @@ metadata:
     app: kuberhealthy
     prometheus: prometheus
   name: kuberhealthy
+  namespace: kuberhealthy
 spec:
   jobLabel: component
   selector:
--- a/deploy/kuberhealthy-prometheus-operator.yaml
+++ b/deploy/kuberhealthy-prometheus-operator.yaml
@@ -12,8 +12,7 @@ spec:
     singular: khcheck
     kind: KuberhealthyCheck
     shortNames:
-      - khc  
-      
+      - khc
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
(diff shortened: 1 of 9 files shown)


integrii:
Thanks for the PR to address the permissions issue!

Unfortunately these specs are automatically generated from the helm chart here: https://github.com/Comcast/kuberhealthy/tree/master/deploy/helm/kuberhealthy

Would you mind dropping your changes in the helm templates instead of on these flat files, then running the `create-flat-files.sh` script [here](https://github.com/Comcast/kuberhealthy/blob/master/deploy/helm/create-flat-files.sh).?

qqshfox:
@integrii Updated. Thought the chart was outdated and not maintained.

janavenkat:
@integrii, please review

deploy/kuberhealthy-prometheus-operator.yaml
```diff
@@ -301,9 +328,6 @@ metadata:
   name: kuberhealthy
   namespace: kuberhealthy
   annotations:
-    prometheus.io/scrape: "true"
-    prometheus.io/port: "80"
-    prometheus.io/path: "/metrics"
```
integrii: Why drop the prometheus scrape annotations here?  I would prefer to scrape the service for the global state of checks rather than checking each pod individually; having them overwrite each others information.

integrii:
The chart at `helm/charts/stable` is outdated, but the one in this project is treated as the master for all specs.  We generate the flat files in this repo from it rather than maintaining them individually.  Sorry for the miscommunication there. @qqshfox 

I just had one quick comment on the removal of scrape annotations on the service.  Other than that, I think this is worth merging.  If something goes wrong, we can do further tweaks then.

Thanks again for going through all this RBAC @qqshfox!

deploy/kuberhealthy-prometheus-operator.yaml
qqshfox: There's a `ServiceMonitor` in `kuberhealthy-prometheus-operator.yaml` already. These annotations had no effect here. We only need these in `kuberhealthy-prometheus.yaml`.

vishalud:
Thank you !This works as expected !

Issue #303, opened 2020-02-10
Kuberhealthy service account can't evict pods


integrii:
Here are the logs from kuberhealthy wherein it was unable to evict pods:

```
 ~  kaorb logs kuberhealthy-6d4ccbc4bd-hqnzk | grep deployment-1581372373                                                                                                                                                                                                                               master  2:31:21P 
time="2020-02-10T22:06:20Z" level=info msg="5b9ae1fa-bab2-4b0c-853c-1b2858a7d102 kuberhealthy/deployment: [Creating external checker pod named deployment-1581372373]"
time="2020-02-10T22:06:20Z" level=info msg="5b9ae1fa-bab2-4b0c-853c-1b2858a7d102 kuberhealthy/deployment: [Check deployment created pod deployment-1581372373 in namespace kuberhealthy]"
time="2020-02-10T22:06:24Z" level=info msg="5b9ae1fa-bab2-4b0c-853c-1b2858a7d102 kuberhealthy/deployment: [External check pod is running: deployment-1581372373]"
time="2020-02-10T22:06:24Z" level=info msg="5b9ae1fa-bab2-4b0c-853c-1b2858a7d102 kuberhealthy/deployment: [Waiting for pod status to be reported from pod deployment-1581372373 in namespace kuberhealthy]"
time="2020-02-10T22:21:15Z" level=info msg="5b9ae1fa-bab2-4b0c-853c-1b2858a7d102 kuberhealthy/deployment: [Evicting up any running pods with name deployment-1581372373]"
time="2020-02-10T22:21:15Z" level=info msg="5b9ae1fa-bab2-4b0c-853c-1b2858a7d102 kuberhealthy/deployment: [error when trying to cleanup/evict checker pod deployment-1581372373 in namespace kuberhealthy: pods \"deployment-1581372373\" is forbidden: User \"system:serviceaccount:kuberhealthy:kuberhealthy\" cannot create pods/eviction in the namespace \"kuberhealthy\"]"
```

PR #315, opened 2020-02-19
regenerate flat spec files

integrii:
- flat spec files regenerated to include recent fixes to daemonset check role binding and eviction permission


integrii:
should fix #303

Issue #293, opened 2020-01-26
Kuberhealthy memory leak


cagdascirit:
**Describe the bug**
Kuberhealty master pod is leaking memory and when it hit its limit value it is being killed. Kuberhealthy follower pods are not suffering from this issue. This can be related to khc->podcreation->khsupdate cycle.

**Steps To Reproduce**
- Deploy kuberhealty deployment (with at least one replica) and khc objects
- Put 12 external healthchecks (for us it is 13 in total with dns-check)

**Expected behavior**
Kuberhealthy master and follower pods should not leak memory and keep going stabilized in memory usage.

**Screenshots**
[Here memory request is 300mb, memory limit is 1gb]
![Screenshot 2020-01-26 at 21 14 34](https://user-images.githubusercontent.com/2715555/73141234-23df8a00-4082-11ea-9482-624f77b57d76.png)

**Versions**
 - Cluster OS: Container Linux by CoreOS 2303.3.0 (Rhyolite)
 - Kernel Version: 4.19.86-coreos
 - Docker  Version: 18.6.3
 - Kubernetes Version: 1.14.8
 - Kuberhealthy Release: 2.0.0 (and also tested with 2.1.0rc1 same behaviour)

**Additional context**

integrii:
Whoa!  Thanks for reporting this.  I can't repro in any of our clusters.  We get charts like this:
![image](https://user-images.githubusercontent.com/98695/73196591-b749a700-40e4-11ea-9e45-a3e71174a338.png)


Can you paste me your `khcheck` definitions?  Do you see any errors at the console of either of the Kuberhealthy pods showing a memory leak?

cagdascirit:
"Do you see any errors at the console of either of the Kuberhealthy pods showing a memory leak?"
No, only some warnings like below on leader and follower pods:

> level=warning msg="was unable to find calling pod with remote IP X.X.X.X while watching for duration"

cagdascirit:
We have in total of 13 khchecks deployed. Your grafana screenshot window is too short to judge if there is a leak or not. But what I understand from your statements that on your clusters memory usage is consistent even in bigger time windows right? weird..

integrii:
I realized after posting that my chart was indeed worthless.  Here is a better one for proofsies.

![image](https://user-images.githubusercontent.com/98695/73417586-71036c00-42cd-11ea-9433-110ba1057bcb.png)

However, while looking for a bad pod, I did find one with a memory leak!

[...]

PR #317, opened 2020-02-20
Bug/293

integrii:
- fixes for memory leaks related to watches never ending

image


cagdascirit:
https://github.com/Comcast/kuberhealthy/issues/293#issuecomment-590119623 ):

integrii:
merging this for now even though it does not fix #293

PR #318, opened 2020-02-21
apiGroup typo

integrii:
- fixes typo in helm chart that causes incorrect RBAC on daemonset checker

### DIFF
--- a/deploy/helm/kuberhealthy/templates/khcheck-daemonset.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-daemonset.yaml
@@ -48,7 +48,7 @@ rules:
   - apiGroups:
       - ""
       - extensions
-      - app
+      - apps
     resources:
       - daemonsets
       - pods


qqshfox:
Tested locally and this worked for me as well.

zeleena:
This works for the helm chart but ideally this should be changed everywhere this is mentioned, and the typo still exists in the following flat files:

- https://github.com/Comcast/kuberhealthy/blob/master/deploy/kuberhealthy.yaml#L216
- https://github.com/Comcast/kuberhealthy/blob/master/deploy/kuberhealthy-prometheus.yaml#L246
- https://github.com/Comcast/kuberhealthy/blob/master/deploy/kuberhealthy-prometheus-operator.yaml#L246

joshulyne:
also, could we get rid of the apigroup "extensions" now?

integrii:
I am going to attempt to add a trigger for github to rebuild the flat files on each master merge...

integrii:
For now, lets merge this.  Then, lets regen the flat files.  The automatic flat file regen github action is ready for initial tests in #327

integrii:
@joshulyne I don't think we can drop the `extensions` api version yet.  

The Kubernetes 1.11 documentation still lists custom resources as using it:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#customresourcedefinition-v1beta1-apiextensions-k8s-io

PR #321, opened 2020-02-24
Add envvar to customize reporting URL in helm chart

zeleena:
The `KH_EXTERNAL_REPORTING_URL`, the service URL for the status page, can be customized based on the code here: https://github.com/Comcast/kuberhealthy/blob/3720681d1bd87604e95610a0d49241cc296738c9/cmd/kuberhealthy/main.go#L63

To be able to use this feature in the provided helm chart, this value should be able to be set in the values file. This PR adds just this environment variable to the Deployment container. Only this environment variable was added in this PR to constrain what environment variables can be set.

### DIFF
--- a/deploy/helm/kuberhealthy/templates/deployment.yaml
+++ b/deploy/helm/kuberhealthy/templates/deployment.yaml
@@ -62,6 +62,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          {{- if .Values.deployment.env.KH_EXTERNAL_REPORTING_URL }}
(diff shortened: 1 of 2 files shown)


zeleena:
Thanks @joshulyne ! I've actually submitted a feature request https://github.com/Comcast/kuberhealthy/issues/322 too, which is similar to this PR but I thought it might be better to point out first in case it would belong to a larger, ongoing effort.

Also, thanks for the review! Feel free to merge when you see fit, as I don't have permissions to 🙂

integrii:
LGTM.  Thanks @zeleena !

PR #335, opened 2020-03-02
working on push path filtering

integrii:


integrii:
This works. See PR #335 
I also left a filter open for triggers from branch `flat-file-gen` so that we can test changes to the generation without the master branch being used.

Issue #333, opened 2020-03-01
Least privilege permissions for reaper


omerlh:
**Describe the feature you would like and why you want it**
Reaper cron job currently needs admin permissions on the cluster, and for example, try to list pods at the cluster scope. This is very sensitive permission to grant for this job, I wish to limit it to a namespace scope. I did that manually for kuberhealthy itself by changing the RBAC roles, but for reaper, it requires code change.

integrii:
Thanks for reporting this.  Because Kuberhealthy is a cluster-wide operator that may be making checks in any namespace, the reaper also tries to look for things to delete in all namespaces. If that is not the desired behavior, then I suggest the following:

- identify the code path (api calls) that won't work without cluster-wide bindings
- create an alternate code path path that works when you only have bindings for a single namespace
- add an option (or options) via environment variable that switches the reaper to run in the single namespace mode

integrii:
looks like the fix PR was merged - we just need to push the image somewhere, then set it in the helm chart's spec for the reaper and let the bot regen the flat files...

joshulyne:
Addressed with pr #340 and #338

PR #338, opened 2020-03-04
Make check reaper have single namespace privilege

joshulyne:
Addresses #333


integrii:
Next steps: push the image and update the helm chart.

PR #342, opened 2020-03-10
Resource quotas

jonnydawg:
Makes a check for resourcequotas.

By default checks if CPU or memory are over a threshold (0.9 or 90%) for specified namespaces.


cmd/resource-quota-check/README.md
```diff
@@ -0,0 +1,78 @@
+## Resource Quotas (CPU & Memory)
+
+This check tests if namespace resource quotas `CPU` and `memory` are under a specified threshold or percentage. Namespaces that utilize a lot of `CPU` or `memory` resources can sometimes run into an issue where controllers (_i.e. deployment or replica controllers_) are unable to schedule pods due to insufficient `CPU` or `memory`.
+
+This check lists all namespaces in the cluster and checks if each resource (`CPU` and `memory`) are at an ok percentage.
+
+This check can be configured to use either a `blacklist` or a `whitelist` of namespaces, allowing you to explicitly target or ignore specific namespaces. This can be changed with the environment variables `BLACKLIST` and `WHITELIST` which take `"true" or "false`. This check assumes the use of a `BLACKLIST` on check runs; if both `BLACKLIST` and `WHITELIST` options are enabled, `BLACKLIST` will take precedence. The default lists are:
```
integrii: What about refactoring this so that both a blacklist or whitelist can be used in any way the user likes?

Here is the matrix of functionality:
- if blacklist is specified, and whitelist is not, then we simply operate on a blacklist
- if blacklist is specified, and whitelist is also specified, then we operate on the whitelist unless the item is in the blacklist
- if blacklist is not specified, but whitelist is, then we operate on a whitelist
- if neither a blacklist or whitelist is specified, then all namespaces are targeted

This may sound pointless, but perhaps someone may use it programmatically in the future and it will be easy for them.  This should also simplify our code, because we can simply ask a few boolean questions to implement all functionality, rather than operating on some mutating state machine that has various types of functionality.

cmd/resource-quota-check/README.md
```diff
@@ -0,0 +1,78 @@
...
+```yaml
+---
+apiVersion: comcast.github.io/v1
+kind: KuberhealthyCheck
+metadata:
+  name: resource-quota
+  namespace: kuberhealthy
+spec:
+  runInterval: 1h
+  timeout: &resource_quota_check_timeout 2m
```
integrii: Nice use of repeated nodes 🔥

cmd/resource-quota-check/README.md
jonnydawg: Currently only 1 and 3 are implemented:
`if blacklist is specified, and whitelist is not, then we simply operate on a blacklist`
`if blacklist is not specified, but whitelist is, then we operate on a whitelist`

I think here, I would need to add another case to my `switch` statement on `line: 110` in `resources.go`

jonnydawg: In theory, would it be bad to run two of these checks to achieve the same outcome? 
Have a `resource-quota-blacklist` and `resource-quota-whitelist` check that operate on blacklists and whitelists respectively -- thoughts?

Makefile
```diff
@@ -79,16 +79,24 @@ buildHTTPCheck:
 pushHTTPCheck:
 	docker push quay.io/comcast/http-check:1.0.0
 
+resourceQuotaCheck: buildResourceQuotaCheck pushResourceQuotaCheck
+
+buildHttpContentChecker:
+	docker build -t jdowni000/http-content-check:v1.1.1 -f cmd/http-content-check/Dockerfile .
+
```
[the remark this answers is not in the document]
integrii: #344 has been completed.  Check out the new Makefile structure.  Each directory in `cmd` has a `Makefile` and `Dockerfile`.  Add one to this project, too.

cmd/resource-quota-check/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.14 AS builder
+ADD . /build
+WORKDIR /build/cmd/resource-quota-check
+RUN CGO_ENABLED=0 go build -v
+
+FROM golang:1.14.0-alpine
```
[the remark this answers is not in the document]
integrii: actually, we can't use `alpine` anymore.  Check out issue #349

cmd/resource-quota-check/main.go
```diff
@@ -0,0 +1,141 @@
...
+		return
+	}
+	log.Infoln("Kubernetes client created.")
+
+	// Catch panics.
+	var r interface{}
+	defer func() {
+		r = recover()
+		if r != nil {
+			log.Infoln("Recovered panic:", r)
```
integrii: Very nice.  We should have these in more of our checks.

cmd/resource-quota-check/main.go
```diff
@@ -0,0 +1,141 @@
...
+func listenForInterrupts(ctx context.Context) {
+
+	// Relay incoming OS interrupt signals to the signalChan.
+	signal.Notify(signalChan, os.Interrupt, os.Kill, syscall.SIGTERM, syscall.SIGINT)
+	sig := <-signalChan // This is a blocking operation -- the routine will stop here until there is something sent down the channel.
+	log.Infoln("Received an interrupt signal from the signal channel.")
+	log.Debugln("Signal received was:", sig.String())
+
+	log.Debugln("Cancelling context.")
+	ctxCancel() // Causes all functions within the check to return without error and abort. NOT an error
```
integrii: normally this is passed into the function, or you can run `ctx.Cancel()` I think.

cmd/resource-quota-check/resources.go
```diff
@@ -0,0 +1,204 @@
...
+			log.Infoln("Skipping", namespace, "namespace")
+			return
+		}
+		examineResouceQuotasForNamespace(namespace, quotasChan)
+		return
+	// By default, look at all namespaces.
+	default:
+		examineResouceQuotasForNamespace(namespace, quotasChan)
+		return
+	}
```
integrii: I bet you can do this much simpler.  Try using a small number of `if` statements and using `return` where appropriate to skip the namespace.

Try getting rid of your `whitelistOn` and `blacklistOn` globals entirely if you're up for it.

If you don't feel like the refactor, I think this approach is fine for a first release.

PR #345, opened 2020-03-11
fix `safe-to-evict` annotation

qqshfox:
should be added in `podTemplate` instead of in `DaemonSet`


integrii:
Makes perfect sense.  Thanks for your continued contributions!

PR #352, opened 2020-03-12
github actions created for each CMD dir

jdowni000:
created build-latest YAML files for each `CMD` dir


jdowni000:
updated the names of the github action yaml files @integrii

integrii:
Looks like a good start - let's watch these actions and make sure that builds are working as expected going forward, though.

PR #360, opened 2020-03-16
allow to customize envs

qqshfox:


integrii:
Looks good.  Thanks as always @qqshfox !

PR #376, opened 2020-03-20
Dockerhub

jdowni000:
Updated the images for all checks excluding kuberhealthy since it is still broken with the new image.


jdowni000:
Updated Helm `values.yaml` file to reflect to Dockerhub images with correct tags

Issue #366, opened 2020-03-17
ownerReferences added to all object


NissesSenap:
**Describe the feature you would like and why you want it**
Would love to see ownerReferences in all objects that is created by the kuberhealthy crd:s to make it easier to track what creates what and where to debug for new users.

**Additional context**
This is standard within k8s itself, for example see the check-reaper pod, you will see a ownerReference to the job object.

I think this will take some time, so but i don't think it's hard.

joshulyne:
Hmm I believe in the checker pods, we have `kuberhealthy_check_name=pod-status` which is referencing the khcheck crd `pod-status` -- is this what you're referencing?

integrii:
I think @NissesSenap means he wants us to implement [owner reference metadata](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#ownerreference-v1-meta).  I think this is a good idea to do!  Perhaps good for a first time contributor?

PR #378, opened 2020-03-22
Add ownerRefernce to external check creation

NissesSenap:
This doesn't work. Need input.
This is my initial stab at #366


NissesSenap:
Can someone point me in the right direction? 
My go skills is still low and so i get lost within channels and interfaces.

I wanted to send the the UID together with name etc. that is needed for the ownerReference meta data to the checks created but from the kuberhealthy I can't manage to find how to get the data from the CRD and then push it towards the objects straight away.

Instead I tried to solve it by pulling the data from the CRD as seen in my PR. 
First of all it doesn't work due to missing info (could do some strange for loops to make a educated guess) and it would give unneeded access to the SA and when i write this comment I realize updating the final object from itself sounds like a dumb idea... :) 

Any way if someone can point me in the correct direction I think i should be able to solve it.

integrii:
Hey @NissesSenap - Thank you for hacking on this!  Unfortunately, I think the current approach may not be quite right.

I think the right place to do this is around [here](https://github.com/Comcast/kuberhealthy/blob/aab7c6f8d8f4acd9e3c8e21c69298dfc74fbe653/pkg/checks/external/main.go#L1028).  Basically we want to go into the `p` variable here and set the right Owner metadata information.  According to the docs, the metadata should be at `p.ownerReferences` (or something similar) and should be of type `OwnerReference`. 

Ultimately, I think that when a pod is programmatically created (anywhere) we should include this owner information.  It's really good stuff and I am happy you brought it up.  Making this one change would implement this owner information for all checker pods created by Kuberhealthy... Which is awesome.

The next step would be to go through each checker pod and ensure that anytime it creates a pod or daemonset or whatever, it always specifies the right `OwnerReference` object that points back to the checker pod.

NissesSenap:
Man I don't know how many hours i was looking for something like createPod() but ofc i should have looked in the pkg and not in the cmd/kuberhealthy...
I did a force push so i removed the old code from my branch so i didn't have to recreated the PR.

I'm not sure I found the correct path this time ether but it's definitely closer.
Now i can see a ownerRefernece in the created objects but they point back to there own controller which makes sense but i guess i hoped it would append to the list and be able to still refer back to the kuberhealthy deployment.

Or I still haven't been able to wrap my head around it ^^ 
Anyway need to do some other things now but please have a look and I will try to get some more time for this soon.

integrii:
Thanks @NissesSenap !  I will take a look at this one tomorrow...  I didn't get enough time to investigate your PR today.

integrii:
Have you by any chance built/tested this one yet on one of your own clusters?  Our tests are almost non-existent due to how hard it is to spin up, tear down, and instrument a cluster for testing.

NissesSenap:
Yes I have tried it but it dosen't do the trick the ownerRef gets overwritten by the deployment, I will need to take a look at it again, but currently i need to prioritize other things. 
If you or someone else have time to look at it please go for it.

I run Kubernetes Version: v1.16.2 (Openshift version 4.3.0)

pkg/checks/external/main.go
```diff
@@ -119,6 +123,7 @@ func New(client *kubernetes.Clientset, checkConfig *khcheckcrd.KuberhealthyCheck
 		OriginalPodSpec:          checkConfig.Spec.PodSpec,
 		PodSpec:                  checkConfig.Spec.PodSpec,
 		KubeClient:               client,
+		crdUID:                   checkConfig.GetObjectMeta().GetUID(),
```
joshulyne, line 126: seems like instead of the khcheck crdUID, you want to reference the kuberhealthy deployment UID? im not sure we have something in the code that gets the kuberhealthy deployment UID... @integrii ? 

may need to add a function somewhere that finds the kuberhealthy deployment and gets the UID from it?

integrii: I would think that the external checker pods that get spun up would have an owner reference that points to the kuberhealthy pod that spawned them (not the deployment or khcheck).  This way, if the kuberhealthy instance that is orchestrating the check disappears for some reason, the checker pods it was orchestrating (and had stateful data for) are also removed.

I think this means we can probably drop all the `crdUID` property stuff and just simply put in the hostname of the kuberhealthy instance that spawns the pod in the owner reference.

NissesSenap: So I will finally take a third stab at this and I will do as you say just point back at the kuberhealthy pod as the owner. I hopefully have something to push later today.

pkg/checks/external/main.go
```diff
@@ -1026,9 +1026,40 @@ func (ext *Checker) createPod() (*apiv1.Pod, error) {
...
+	// Get the podspec of kuberhealthy pod
+	kHealthyPod, err := ext.getKuberHealthyPod(kPodName)
+	if err != nil {
+		ext.log("Unable to find kuberhealthy pod", err)
+	} else {
+		// Sets OwnerRefernces on all checker pods
+		p.OwnerReferences = []metav1.OwnerReference{
+			{
+				APIVersion: "v1",
+				Kind:       "Pod",
```
NissesSenap, line 1044: I tried to find a way to not hardcode the APIVersion and Kind but I was unable to do so.
I'm sure there is a nice way through the k8s core api, if you know it please point it out.

NissesSenap:
I cleaned the CRD stuff as pointed out and put it towards the kuberhealthy master pod.
I seen it work in my own setup, if you want to give it a try you can use:  quay.io/nissessenap/kuberhealthy:2.1.3

NissesSenap:
Was thinking about remaking my function a bit and make it global so the other cmd could use it to reference to them self thus getting a ownerReference on the DS/deployment etc that they create.

But the none of the checks currently just imports comcast/kuberhealthy/v2/pkg/checks/external
so it dosen't make sense. Not that the same function is hard to write but boring to have the exact same code in all the checks.
Could add it in kubeClient but it dosen't feel super natural. 
Could create a util or something like that?
All we need is a KubeClient and namespace and those two things is something the checks always have.
Do you have a good idea on where to put a globally scope function like this?

pkg/checks/external/main.go
NissesSenap: Tried with apiv1.Pod{}.APIVersion but this just returns an empty string

joshulyne:
Hmm I've been looking through that too, where would be a good place for this. So I know all checks import the external/checkclient package since it provides functions for external checks to report back to Kuberhealthy. I'm thinking that could be a spot, but its definitely not perfect. I would say there for now, unless others have an issue with that. @jonnydawg? @integrii is on paternal leave 🎉 rn so he may be slower to respond :P

Both @jonnydawg and I can try importing that function in the daemonset and deployment check once you have it implemented so we can start adding ownerReferences to those checks!

pkg/checks/external/main.go
```diff
@@ -1026,9 +1026,40 @@ func (ext *Checker) createPod() (*apiv1.Pod, error) {
...
+// getKuberHealthyPod gets the pod spec of kuberhealthy operator
+func (ext *Checker) getKuberHealthyPod(kPodName string) (*apiv1.Pod, error) {
+	podClient := ext.KubeClient.CoreV1().Pods(ext.Namespace)
+	kHealthyPod, err := podClient.Get(kPodName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return kHealthyPod, nil
+}
+
```
joshulyne, line 1062: wondering if this could all be combined to a single function `addOwnerReference` where it takes in the kube client and namespace (like you mentioned with making this more of a global function) and returns an ownerReference.

NissesSenap: Got an issue with circular imports so i created a separate package that i called utils and made it too two functions where ownerRef calls on the get pod one. Please have a look at the latest commit

NissesSenap:
Due to circular imports I created a new package that i called uitls, maybe not the best name.

It won't be any issue for you to call this package from the different checks.
If you don't like this solution we can ofc have the same code base in two places.

Hopefully you think this looks okay. I can also have a mock fest and try to write some test for my very small package.

pkg/checks/external/checkclient/main.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/Comcast/kuberhealthy/v2/pkg/checks/external/status"
 )
 
+const (
+	podAPIVersion = "v1"
+	podKind       = "Pod"
+)
+
```
joshulyne, line 23: is there a reason this is needed here? :)

pkg/checks/external/main.go
```diff
@@ -1026,6 +1028,15 @@ func (ext *Checker) createPod() (*apiv1.Pod, error) {
...
 	ext.addKuberhealthyLabels(p)
 
+	// Get ownerRefernece for the kuberhealthy pod
+	ownerRef, err := util.GetOwnerRef(ext.KubeClient, ext.Namespace)
+	if err != nil {
+		ext.log("Unable to get the ownerRef from pod", err)
+	} else {
+		// Sets OwnerRefernces on all checker pods
+		p.OwnerReferences = ownerRef
+	}
```
joshulyne, line 1038: Hmm after @jonnydawg and I look through this a bit, I believe we want to return the error instead of just logging it

joshulyne:
Overall, I think this looks good! I think as long as we bubble up the error we should be good to go! @jonnydawg and I should be able to import this package to the ds and deployment checks 👍 thanks so much!

go.mod
```diff
@@ -16,13 +16,15 @@ require (
...
 	github.com/gogo/protobuf v1.3.0 // indirect
+	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
 	github.com/google/uuid v1.1.1
 	github.com/gophercloud/gophercloud v0.1.0 // indirect
 	github.com/influxdata/influxdb1-client v0.0.0-20190402204710-8ff2fc3824fc
 	github.com/integrii/flaggy v1.2.2
 	github.com/pkg/errors v0.8.1
 	github.com/pkg/sftp v1.10.1 // indirect
-	github.com/sirupsen/logrus v1.4.0
+	github.com/prometheus/common v0.9.1
```
jonnydawg, line 26: Do we need this?

NissesSenap: Nope, me being unused to go package management and these damm auto updates.
I will remove it in the next commit.

pkg/checks/external/checkclient/main.go
NissesSenap: Nope, me that forgot to clean them after I moved the function. 
Thanks for spotting it.

pkg/checks/external/main.go
NissesSenap: My thought was that it isn't worth not creating the pod just because you can't find the ownerRef so that's why I did a log instead of a return.

But I have updated the code to return the error and removed the else.

pkg/checks/external/util/main.go
```diff
@@ -0,0 +1,44 @@
...
+		{
+			APIVersion: podAPIVersion,
+			Kind:       podKind,
+			Name:       podSpec.GetName(),
+			UID:        podSpec.GetUID(),
+		},
+	}, nil
+}
+
+func getKuberHealthyPod(client *kubernetes.Clientset, namespace, podName string) (*apiv1.Pod, error) {
```
joshulyne, line 37: nitpicky comments: 
KuberHealthy --> Kuberhealthy :) 
Also, a comment description for this function would be good too!

joshulyne:
Also, tested this locally, and it worked well! When I delete the kuberhealthy instance responsible for all the external checks, all the completed external checks were deleted -- super helpful for cleanup purposes :)

Issue #237, opened 2019-10-24
Create a Javascript module for reporting status to Kuberhealthy and an example check for it


integrii:
It would be nice to have a Javascript module and example `khcheck` for Javascript developers to start with when writing their own checks.

integrii:
docs for making your own check client: https://github.com/Comcast/kuberhealthy/blob/2.0.0/docs/EXTERNAL_CHECK_CREATION.md

jonnydawg:
new doc should be at:
https://github.com/Comcast/kuberhealthy/blob/master/docs/EXTERNAL_CHECK_CREATION.md

jonnydawg:
Also looking at OAS for this for clients in multiple languages

jonnydawg:
This has been created, but the client is intermittently posting status. 

![image](https://user-images.githubusercontent.com/12464614/78174203-be0ad800-740d-11ea-9e1f-c273933634bd.png)

jonnydawg:
Going to take a look at async / Promises

jonnydawg:
addressed by @gwolf3 in #476

jonnydawg:
There is a sample javascript check under the `clients/` directory. Documentation about using a javascript client has been updated to include @gWOLF3 NPM package. [here](https://github.com/Comcast/kuberhealthy/blob/master/docs/EXTERNAL_CHECK_CREATION.md) and [here](https://github.com/Comcast/kuberhealthy/blob/master/clients/js/README.md)

PR #387, opened 2020-03-26
Client js

jonnydawg:
Implements an external check client for kuberhealthy in JavaScript.

Comes with example Dockerfile and check.


jonnydawg:
Addresses #237

clients/js/example/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM node 
+COPY package*.json ./
+RUN npm install
+
+COPY . .
```
integrii: It looks like these two copies duplicate eachother?  Why not just do a `COPY . .` first here?

integrii:
Can you also add instructions to the main README.md and the EXTERNAL_CHECK_CREATION.md readme instructs people to look in the new clients directory for non-go clients?

PR #403, opened 2020-04-02
create docs for deploy directory

integrii:
documentation for the deploy dir files to make install less confusing


2infinitee:
@integrii I think it's pretty straight forward how you have it at the moment. I think it can be more consistent and organized. 

As for the Helm portion of it, it's interesting how you are using Helm. I think it would be easier to provide a`values.yaml` files for each configuration (w/ prometheus operator, prometheus, or no prometheus). Can tell you more about it.

PR #409, opened 2020-04-08
go mod download for faster Dockerfile

NissesSenap:
go mod download will help to download all dependencies before coping over the code. This won't help allot in your local environment but if you run this in a CI/CD pipeline you will see a increase if you run the build multiple time at the same server due to docker caching.

Since go creates a bin and the COPY command sets the owner to the current user we don't have to do any chown until the bin is ready.
I assume that dockerhub should still be happy with the dockerfile since we are using USER in the final container.

I only did this update in one file, if you think this is a good idea i will do it in the rest of them.

### DIFF
--- a/cmd/ami-check/Dockerfile
+++ b/cmd/ami-check/Dockerfile
@@ -1,10 +1,14 @@
 FROM golang:1.13 AS builder
-RUN groupadd -g 999 user && \
-    useradd -r -u 999 -g user user
-COPY --chown=user:user . /build
+WORKDIR /build
+ADD go.mod go.sum /build/
+RUN go mod download
+
+COPY . /build
 WORKDIR /build/cmd/ami-check
 ENV CGO_ENABLED=0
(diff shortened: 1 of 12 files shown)


NissesSenap:
@integrii  please have a look and see what you think.

integrii:
Looks fine to me as long as the build still works with this new `Dockerfile`.  You're a real docker wizard!

Thank you for looking at Kuberhealthy so seriously. We appreciate all your contributions a lot.

NissesSenap:
Thank, glad that I can help with something I feel comfortable doing.

I have updated the rest of the Dockerfiles and I hope I managed to get them all right.
I only had time to build kuberhealthy and daemonset but I will try the rest of them tonight.

joshulyne:
I've also been able to test some of the Dockerfiles you changed, and they lgtm! Haven't been able to test all of them though -- will try and do that soon!

PR #412, opened 2020-04-14
MOD:fix client return fmt.Errorf

calmkart:
fix "%s" with two variable
just a small fix.


calmkart:
@integrii 
PTAL, thanks

integrii:
Good catch!

Issue #373, opened 2020-03-19
Latest kuberhealthy image from docker hub has blank status page


integrii:
The status page on the latest Kuberhealthy build `:latest` has a blank `khstate` page. 

It appears to be unable to use its reflector for `khstate` resources:

```
E0318 20:54:04.973308       1 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *khstatecrd.KuberhealthyState: no kind "KuberhealthyStateList" is registered for the internal version of group "comcast.github.io" in scheme "pkg/runtime/scheme.go:101"
```

integrii:
Looks like the reflector for `khstate` is having issues with the `khstate` crd code.  This code changed somewhat recently in K8s so we need a reintegration when it comes to declaring the CRD's codec and listing abilities.

integrii:
current strategy is to look into re-creating our CRD packages with the [operator SDK](https://github.com/operator-framework/operator-sdk)...

PR #415, opened 2020-04-14
Revert "modify version in helm, package v.2.1.2, and add to existing index.yaml"

joshulyne:
Reverts Comcast/kuberhealthy#410


joshulyne:
reverting changing build to 2.1.2 since we can't create new kh builds until #373 is fixed

Issue #353, opened 2020-03-12
Possible race between khstate and checks reporting in


integrii:
It may be that we have a race condition wherein checks are able to report in before the `khstate` for them reflects their current run IDs.  This results in the check being served a `400` because it does not appear to have the correct check run id.

This screen shot has some more brain storming
![image](https://user-images.githubusercontent.com/98695/76473822-e6775780-63b6-11ea-9865-fcbef338c047.png)

I suggest that we put some code in that validates that the `khstate` object has the expected UUID before creating the checker pod.  This would ensure that the checker pod can't report in before the proper `khstate` uuid is reflected.

integrii:
The check needs added right after the code here: https://github.com/Comcast/kuberhealthy/blob/master/cmd/kuberhealthy/kuberhealthy.go#L1014

PR #419, opened 2020-04-15
Github Actions e2e test with kind

NissesSenap:
To get some of the tests to work we need a k8s cluster. Played around a bit with Kind and got a super simple e2e to work using bash...

When #353 is solved we could start to trigger the tests maybe for each PR or at least for each tag before we do a push to docker.

This PR is more of a discussion point but I hope you want to use it or something like it + I wanted to learn how GitHub actions worke :)


NissesSenap:
It seems like the initial operator deploys two sets of each check. 
DS and DNS seems to work fine but deployment not so much.
In KSH I get:
    - services "deployment-svc" already exists

So due to the double pods that get created they get in each others way...
Not really sure how to solve that in a smooth way, can ofc do so more ugly bash scripting but it wasn't really the point of this PR ^^.

NissesSenap:
Or i get:
- deployments.apps "deployment-deployment" not found

You can find the logs in my fork -> branch actions -> actions -> latest build if you want to check it out :)

joshulyne:
Wow, this looks awesome -- will definitely check this out and test for ourselves! Thanks so much ❤️ @NissesSenap

.ci/e2e.sh
```diff
@@ -0,0 +1,74 @@
...
+name=kuberhealthy
+
+# Create namespace
+kubectl create namespace $NS
+
+# Sometimes the kuberhealthy resources get's created...
+sleep 2
+
+# Create kuberhealthy crd etc.
+kubectl create -f .ci/kuberhealthy-e2e.yaml
```
joshulyne, line 19: Hmm I'm wondering if we could use helm to install kuberhealthy instead -- this could make the e2e test more configurable cause then instead of having to add `.ci/kuberhealthy-e2e.yaml` file, its just a `values.yaml` file that can be configured to enable or disable certain external checks while running e2e testing. I know the default kuberhealthy configuration only enables those 3 checks, but it'd be nice for certain PRs that are making changes to external checks to add them to the e2e testing. Not sure how configurable github actions can be for that :woman_shrugging:

.ci/e2e.sh
```diff
@@ -0,0 +1,74 @@
...
+selector=$(kubectl -n $NS get --raw "${selfLink}/scale" | jq -r .status.selector)
+kubectl logs -n $NS --selector $selector
+
+# Verify that the khc went as they should.
+for i in {1..60}
+do
+    khsCount=$(kubectl get -n $NS khs -o yaml |grep "OK: true" |wc -l)
+    cDeploy=$(kubectl -n $NS get pods -l app=kuberhealthy-check |grep deployment |grep Completed |wc -l)
+    cDNS=$(kubectl -n $NS get pods -l app=kuberhealthy-check |grep dns-status-internal |grep Completed |wc -l)
+    cDS=$(kubectl -n $NS get pods -l app=kuberhealthy-check |grep daemonset |grep Completed |wc -l)
```
joshulyne, line 46: If we were to use helm -- there'd need to be a way to grep the `enabled` values for each check to verify the checks completed their runs

joshulyne:
Went through the logs and I do see kuberhealthy provisioning two of each check -- which isn't what it should do :( wondering if thats the reason for the deployment check failure, it shouldn't see two deployment checks running at the same time... I see that you set the kuberhealthy replica to 1 -- any particular reason for that? Also, wondering for this e2e to if it'd be worth deleting the khcheck after it runs successfully or making run intervals longer -- cause I believe we only need the check to run successfully once? Just throwing out ideas here! 

Will continue to investigate further into the duplicate provisioning of checks!

NissesSenap:
Yeah I think the deployment duplicates is the main problem.
My current guess is that the operator automatically triggers a check when being created and then a few seconds later it hits the timing for the schedule but I haven't read the code for this part so I don't know.

The reason why i do 1 kubrehealthy replica is due to that we have limited resources in the GitHub actions environment so the less load we can create the better I thought. My thought was since it's active-passive it shouldn't matter any way.

Was thinking the same thing, As you say one successful one is all that we should care about.
It's probably easier to set long intervals compared to deleting  khchecks.

.ci/e2e.sh
NissesSenap: Hmm I haven't used helm in production for a year or something like that but if I remember correctly helm2 didn't care about if the pods actually starts it just cares if it have performed a apply more or less and a helm ls would still show installed.

I don't think this have changed in helm3, if so great.
So I think we would still have to check if the pods is actually finished a run.

.ci/e2e.sh
NissesSenap: It's a good point, I will play around with helm and provide a update to this PR.

Sadly the AMI check will always be hard to test.
I guess we could create a IAM user in some test environment where we can setup some data, but it's an edge case so let's focus on the low hanging fruit first.


We could trigger to all PR:s and do some ninja script to parse out what files have changed since the last commit and only trigger e2e tests due to those changes but it would be a pain to maintain.

I don't see a big issue with triggering al the e2e tests on all the PR:s, sure each run will take a few minutes longer but I would say it's worth it compared to maintaining strange scripts.
But to make sure that we don't start e2e tests unless we have to I think we should trigger 
go test ./... firsts so if someone sen't in a PR that wasn't possible to build/test we would just cancel the job instead of running all of the tests.

NissesSenap:
There is one major thing that isn't solved.
How should we actually test the latest code?

We need to upload a newly built container for everything to run. We can ofc perform a upload to latest for all containers that you maintain and we just point to latest in the helm chart and make sure that everyone understand that it's latest is unstable and they should always use tag versions or you could do some CI container repo.

I don't have any awesome solution for this but worth thinking about.

.ci/e2e.sh
joshulyne: Yeah, I think ami-test is also environment specific so I'm okay skipping that. 
I agree with e2e only being triggered after we know the PR can pass tests / build!

.ci/e2e.sh
joshulyne: Ignore this comment ^^ since we're going to be running all checks, we'd just have to check that each one successfully runs!

joshulyne:
hmm @NissesSenap thats a good point. Using "latest" could be a temporary solution, but definitely not the safest as users may use it unknowingly. Ideally it'd be nice for every PR (that needs e2e testing) to automatically get its own kuberhealthy build version that we use to run the e2e test against... that would mean having that CI container repo to manage these builds <-- something we should think about, esp if we really want to take advantage of these e2e tests!

joshulyne:
Hmm so when a check gets applied, Kuberhealthy notices the changes applied and takes its time before applying the configurations. From the logs below, you can see when the two checks were applied to when Kuberhealthy enabled the external checks (about 30 secs). I'm curious to see more detailed logs from Kuberhealthy when the checks are first applied to see how the duplicates are happening 🤔 

```
time="2020-04-17T22:16:51Z" level=info msg="channel notify limiter witnessed an upstream message on inChan"
time="2020-04-17T22:16:51Z" level=info msg="channel notify limiter waiting for changes to calm..."
time="2020-04-17T22:17:01Z" level=info msg="channel notify limiter witnessed an upstream message on inChan"
time="2020-04-17T22:17:01Z" level=info msg="channel notify limiter waiting for changes to calm..."
time="2020-04-17T22:17:11Z" level=info msg="channel notify limiter waiting for changes to calm..."
time="2020-04-17T22:17:21Z" level=info msg="channel notify limiter sending a message on outChan"
time="2020-04-17T22:17:21Z" level=info msg="control: Witnessed a khcheck resource change..."
time="2020-04-17T22:17:21Z" level=info msg="control: Reloading external check configurations due to khcheck update"
```

NissesSenap:
I have played around a bit with HELM in my own actions branch and I pushed those changes in to this branch now. I'm using the latest helm chart so we can set how often the runs khchecks should be triggerd.

The problem with deploy is still the same though, all the checks seems to create duplicates when you create kuberhealthy your initial time. I have tried this in my own environment as well and I can see the same thing.
The other applications seem to be resilient to this but sadly not deploy.

You mentioned that you wanted more detailed logs, I haven't looked at it but I guess that you can start in debug mode instead, just have to look how to do it first :)

joshulyne:
Huh, it happens in your local environment as well? I don't think I've seen that before... Do you think you can share the the logs when Kuberhealthy first starts up to when these first checks are created? If you add  `--debug` as a container arg that should enable debug mode!

NissesSenap:
So had some time this morning so I added some extra debugging, please take a look at the logs when you have time @joshulyne 

Also created 2 issues that I found due to this e2e test but I haven't had time to test them locally, hopefully I'm not completely wrong any way.

joshulyne:
@NissesSenap so sorry for the late reply! Went through the kuberhealthy logs of your e2e test that I actually ran locally and I was able to reproduce that error of two sets of khchecks running at initial startup. In the logs, I can see that theres some sort of race condition where khchecks are being reloaded or reconfigured and then not being recognized, so another set of checks are created immediately after: 
```
time="2020-04-29T23:32:18Z" level=info msg="7393c926-55ec-4bb2-9c1f-39f3e5bb9eae kuberhealthy/dns-status-internal: [khstate did not exist, so a default object will be created]"
time="2020-04-29T23:32:18Z" level=info msg="7393c926-55ec-4bb2-9c1f-39f3e5bb9eae kuberhealthy/dns-status-internal: [Creating khstate dns-status-internal kuberhealthy because it did not exist]"
time="2020-04-29T23:32:18Z" level=info msg="1d2db210-4574-44b2-9521-ae79e1c24f95 kuberhealthy/deployment: [khstate did not exist, so a default object will be created]"
time="2020-04-29T23:32:18Z" level=info msg="1d2db210-4574-44b2-9521-ae79e1c24f95 kuberhealthy/deployment: [Creating khstate deployment kuberhealthy because it did not exist]"
time="2020-04-29T23:32:18Z" level=info msg="a28c5535-c9d6-480a-9b2d-9b8ab5c02633 kuberhealthy/daemonset: [khstate did not exist, so a default object will be created]"
time="2020-04-29T23:32:18Z" level=info msg="a28c5535-c9d6-480a-9b2d-9b8ab5c02633 kuberhealthy/daemonset: [Creating khstate daemonset kuberhealthy because it did not exist]"
```
This could be a bug, but I just created a new kuberhealthy build `kuberhealthy/kuberhealthy:v2.1.2-rc.1` that doesn't seems to address the issue above. Let me know if you can test this new build version and if you still see the duplicate khcheck pods!

integrii:
I just want to pop in here and say that this end to end test is super important.  I did not know that KinD existed - very cool!  Thanks to both of you for working on this one.

NissesSenap:
@joshulyne no worries at all, sadly I have had to focus on other tasks at work the last few days.

I triggered the job with your new build and it worked perfectly :100: 
So now the only errors that we got is from pod-restarts and pod-status and it shouldn't be a big problem to fix those.

Now I guess we should decide how to use it ^^

NissesSenap:
@integrii yeah Kind is really cool, it was created by the k8s test sig to be able to test changes done to k8s.

Loading images sounds like a interesting solution.

We have spoken about this problem a bit earlier but how should we actually perform the e2e test preferably without it taking 30+ min to run them?

As mentioned earlier we could point the yaml to the latest build and for each PR that is sen't in we use a github action to rebuild the code that have changed and then we will automatically use that code.
The really bad thing with this solution is of course that we just want to check the code that have changed with this specific PR compared to the rest. + If we would have 2 PR:s close to each other the later one would take the "latest" spot and we would only test that code.

There is ways of solving this but it won't be easy and preferably this solution should be easy to understand and maintain in the long run.

So maybe the best solution is to build all the images for all the PR:s and just load those images in to Kind as you point out.
The bad part about this is of course that it will take time and it's really boring to wait for tests to run. 
But it's probably the easiest solution...


@joshulyne what do you think?

integrii:
I think we should require the test to pass to merge any PR

I also think that we should re-run the test after PRs are merged to ensure that the new master version also passes.  If master fails to build, we should then troubleshoot what went wrong before doing any kind of release...

In an ideal world, it would be nice for users to be able to run the same tests github actions does locally... We might be able to come up with something like [act](https://github.com/nektos/act) to let users iterate on making changes and testing faster.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,31 @@
+name: CI
+
+on:
```
integrii, line 3:
```suggestion
on:
  pull_request:
    branches:
      - master
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,31 @@
...
+    - actions # for testing this build spec
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-go@v1
+      with:
+        go-version: '1.13.3'
+    - uses: actions/checkout@v2
```
integrii, line 17:
```suggestion
    - uses: actions/checkout@v2
    - name: build container
      run: docker build --file cmd/kuberhealthy/Dockerfile --tag kuberhealthy:$GITHUB_RUN_ID .
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,31 @@
...
+      with:
+          version: "v0.7.0"
+
+    - name: Testing
+      run: |
+        kubectl cluster-info
+        kubectl get pods -n kube-system
+        kubectl get nodes
+        echo "current-context:" $(kubectl config current-context)
+        echo "environment-kubeconfig:" ${KUBECONFIG}
```
integrii, line 28:
```suggestion
      run: |
        kubectl cluster-info
        kubectl get pods -n kube-system
        kubectl get nodes
        echo "current-context:" $(kubectl config current-context)
        echo "environment-kubeconfig:" ${KUBECONFIG}
        kind load docker-image kuberhealthy:$GITHUB_RUN_ID
```

.ci/e2e.sh
```diff
@@ -0,0 +1,98 @@
...
+name=kuberhealthy
+
+# Create namespace
+kubectl create namespace $NS
+
+# Sometimes the kuberhealthy resources get's created...
+sleep 2
+
+# Use helm to install kuberhealthy
+helm install -n $NS -f .ci/values.yaml $name deploy/helm/kuberhealthy
```
integrii, line 19:
```suggestion
# the image repository and tag must match the build that just took place
helm install -n $NS --set global.image.repository=kuberhealthy,global.image.tag=$GITHUB_RUN_ID -f .ci/values.yaml  $name deploy/helm/kuberhealthy
```

joshulyne:
@NissesSenap glad the new build worked! haha i didn't add a specific fix to address the above issue ^ but i do know that our old 2.1.0 had a lot of bugs that we've addressed in our master branch. Some things I observed though: 

The errors that I was seeing above `[khstate did not exist, so a default object will be created]` resulted in kuberhealthy to create new checks leads me to believe that it had something to do with the khstate crd. I know that we recently made a bug fix to the khstate crd `https://github.com/Comcast/kuberhealthy/pull/418/files` for this issue: https://github.com/Comcast/kuberhealthy/issues/373 but not sure if it completely addresses these duplicate checks. This error also comes from this function: https://github.com/Comcast/kuberhealthy/blob/master/pkg/checks/external/main.go#L295-L371
where we set the external check UUID. We've recently added this functionality here: https://github.com/Comcast/kuberhealthy/blob/master/pkg/checks/external/main.go#L348-L368 so that we wait to verify that the khstate uuid actually got updated and set properly.

Sorry I can't give you a more concrete example of how we were able to address the duplicate check creation! 

Also, #456 pr is out to address the pod-status and pod-restarts permissions issue and I believe @jdowni000 is working on addressing the `null` error that we were seeing in the pod-restarts report!

integrii:
@NissesSenap When you get time, check out my code suggestions above.  You can approve them with one click if they look good to you.

NissesSenap:
Sorry @integrii for an extremely slow reply. Have had allot to do at work, so didn't have the energy to look at it.
I think your commits looks great so I have merged them.

I guess the main issue is that we will only really test new changes to kuberhealthy and not the underlying  checks . To do that we had to build all the images every time. But let's merge this one in first, see how it runs and see how it goes.
With e2e tests there is always a risk of getting false failed and false positives and we probably need a few runs to find them so I would suggest not making it to advanced from the start.

integrii:
Totally agree.  Here goes nothing!

Issue #400, opened 2020-04-01
use securitycontext with the *-check pods


davidkarlsen:
**Describe the feature you would like and why you want it**
The checker pods fails our pod security policy because they run as priv. user.

**Additional context**
```
  Normal   Pulled     13s                kubelet, alt-eos-g-kwo01  Successfully pulled image "fsdepot.evry.com:8085/kuberhealthy/dns-resolution-check:v1.1.0"
  Warning  Failed     13s (x2 over 13s)  kubelet, alt-eos-g-kwo01  Error: container has runAsNonRoot and image will run as root
```

integrii:
Hey @davidkarlsen!  Thanks for reporting in.  We finally got Kuberhealthy listed as a private repository and people can use `helm install` again.  Glad to see you still around the project.

@jdowni000 is working on adding security scans for Dockerfiles in issue #401 (already in progress but no issue existed).  I also created issue #402 to explicitly track the migration of all images to non-root users.

davidkarlsen:
please note that securityContext != securityScan.
The securityContext controls the UID of the process and would fix the issue I reported. Also I can recommend running the container with securityContext. readOnlyRootFilesystem: https://kubernetes.io/blog/2016/08/security-best-practices-kubernetes-deployment/

NissesSenap:
As a happy OpenShift (OCP) user I thought i would add some extra input, was just going to create a issue about this.
Currently in daemonset-check we have 	runAsUser := int64(1000)
This creates issues in OCP:

Error creating: pods "ds-check-daemonset-1586161692-1586161705-" is forbidden: unable to validate against any security context constraint: [spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1000670000, 1000679999]]

I agree that I think that you should write a securityContext but we should put it ether as 
	runAsUser := int64(1000670001)

Or make it configurable through a env or something like that.
Since it was hardcoded I have built my own version of daemonset-check where I set this high UID and you can try it in:
quay.io/nissessenap/daemonset-check:v2.1.1

davidkarlsen:
as long as it's in a section in the values.yaml the end-user can easily set it to the configuration they want.

integrii:
[...]

PR #423, opened 2020-04-16
Security context additions to khchecks

2infinitee:
Need your thoughts on these changes to the helm charts.
1. Should user and fsgroup be 999 or 1000. I got a bit confused since there was already a security context value for the user.
2. Are the security context where they should be. 

 I added these fields into each of the khchecks pod spec:
```
  podSpec:
    securityContext:
      runAsUser: {{ .Values.securityContext.runAsUser }}
      fsGroup: {{ .Values.securityContext.fsGroup }}
```
and in the container spec:
```
      securityContext:
        runAsUser: {{ .Values.securityContext.runAsUser }}
        fsGroup: {{ .Values.securityContext.fsGroup }}
```
can be adjusted in the `values.yaml` file under:
```
securityContext:
  runAsNonRoot: true
  runAsUser: 999
  fsGroup: 999
  allowPrivilegeEscalation: false
  readOnlyRootFilesystem: true
```

### DIFF
--- a/deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
@@ -9,9 +9,11 @@ spec:
   runInterval: 10m
   timeout: &deployment_check_timeout 15m
   podSpec:
+    {{- if .Values.securityContext.enabled }}
     securityContext:
       runAsUser: {{ .Values.securityContext.runAsUser }}
       fsGroup: {{ .Values.securityContext.fsGroup }}
+    {{- end}}
     containers:
     - name: deployment
       image: {{ .Values.check.deployment.image.repository }}:{{ .Values.check.deployment.image.tag }}
@@ -30,8 +32,8 @@ spec:
         limits:
           cpu: 40m
       securityContext:
-        runAsUser: {{ .Values.securityContext.runAsUser }}
-        fsGroup: {{ .Values.securityContext.fsGroup }}
+        allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation }}
+        readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
       restartPolicy: Never
     {{- if .Values.check.deployment.nodeSelector }}
     nodeSelector:
--- a/deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
@@ -8,17 +8,14 @@ metadata:
 spec:
   runInterval: 10m
   timeout: &deployment_check_timeout 15m
-  securityContext:
-    runAsNonRoot: {{ .Values.securityContext.runAsNonRoot }}
-    runAsUser: {{ .Values.securityContext.runAsUser }}
   podSpec:
+    securityContext:
+      runAsUser: {{ .Values.securityContext.runAsUser }}
+      fsGroup: {{ .Values.securityContext.fsGroup }}
     containers:
     - name: deployment
       image: {{ .Values.check.deployment.image.repository }}:{{ .Values.check.deployment.image.tag }}
       imagePullPolicy: IfNotPresent
-      securityContext:
-        allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation }}
-        readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
       env:
         - name: CHECK_TIME_LIMIT
           value: *deployment_check_timeout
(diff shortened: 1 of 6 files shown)


2infinitee:
https://github.com/Comcast/kuberhealthy/issues/400 tagging issue

2infinitee:
@NissesSenap does this look about right to what you were asking for?

deploy/helm/kuberhealthy/templates/khcheck-dns.yaml
```diff
@@ -9,6 +9,9 @@ spec:
   runInterval: 2m
   timeout: 15m
   podSpec:
+    securityContext:
```
NissesSenap, line 12: Okay to add a if enabled = true for each check?
This way I can disable the security context and let OCP handle it for me.
Else I will have to see the error unable to create with user id 999 okay range is something and then I would have to go in and fix it in helm.

Easy enough to do but It will be a pain if I want to do gitops in the future ^^

NissesSenap: The important part is to be able to enable/disable
    securityContext:
      runAsUser: {{ .Values.securityContext.runAsUser }}
      fsGroup: {{ .Values.securityContext.fsGroup }}

I don't have any need to disable 
        securityContext:
          allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation }}
          readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
But i guess consistency is nice, or connect both of them to the same check.

NissesSenap:
I have tried it and it works with a workaround for me.
If you can implement my review comment as well then I won't even need to do a workaround :).

Thanks for checking @2infinitee

deploy/helm/kuberhealthy/templates/khcheck-dns.yaml
NissesSenap: Can we also add the same if statement to the kuberhealthy deploy file in helm while we are at it :)

2infinitee: Cool I can go ahead and add that in

NissesSenap: Thanks these changes looks good.
But can you add the same if statement to deployment.yaml?

deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
```diff
@@ -30,8 +32,8 @@ spec:
         limits:
           cpu: 40m
       securityContext:
```
integrii, line 34:
```suggestion
{{- if .Values.securityContext.enabled }}
      securityContext:
```

deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
```diff
@@ -30,8 +32,8 @@ spec:
...
-        runAsUser: {{ .Values.securityContext.runAsUser }}
-        fsGroup: {{ .Values.securityContext.fsGroup }}
+        allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation }}
+        readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
```
integrii, line 36:
```suggestion
        readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
{{- end}}
```

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -38,7 +38,10 @@ deployment:
   - /app/kuberhealthy
   # args:
 
+# When enabled equals to true, runAsUser and fsGroup will be
+# included to all khchecks as 999.
 securityContext:
+  enabled: true
```
integrii, line 44:
```suggestion
  enabled: true # if enabled is set to false, securityContext settings will not be applied at all in checker pod custom resources
```

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -38,7 +38,10 @@ deployment:
   - /app/kuberhealthy
   # args:
 
+# When enabled equals to true, runAsUser and fsGroup will be
+# included to all khchecks as 999.
```
integrii, line 42:
```suggestion
# included to all khchecks as specified below.
```

deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
```diff
@@ -26,6 +31,11 @@ spec:
           memory: 15Mi
         limits:
           cpu: 40m
+{{- if .Values.securityContext.enabled }}
+      securityContext:
+        allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation }}
+        readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
+{{- end}}
```
joshulyne, line 38: any reason why the khcheck-deployment requires the `{{- if .Values.securityContext.enabled }}` statement here versus the other checks?

2infinitee: @joshulyne Not sure, but I can add them to all the rest of the other kh checks for consistency as well.

Issue #422, opened 2020-04-16
HELM create variabels runInterval & timeout


NissesSenap:
**Describe the feature you would like and why you want it**
To make it easier to configure the different checks in helm it would be nice to have runInterval & timeout as variabels.

PR #424, opened 2020-04-17
HELM add variabels for runInterval & timeout

NissesSenap:
Fixing #422 also added the possibility to enable & disable checkReaper, I think this will be nice when doing e2e tests.

### DIFF
--- a/deploy/helm/kuberhealthy/templates/check-reaper.yaml
+++ b/deploy/helm/kuberhealthy/templates/check-reaper.yaml
@@ -1,3 +1,5 @@
+{{- if .Values.checkReaper.enabled }}
+---
 apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
@@ -40,3 +42,4 @@ kind: ServiceAccount
 metadata:
   name: check-reaper
   namespace: {{ .Release.Namespace }}
+{{- end }}
--- a/deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
@@ -6,8 +6,8 @@ metadata:
   name: deployment
   namespace: {{ .Release.Namespace }}
 spec:
-  runInterval: 10m
-  timeout: &deployment_check_timeout 15m
+  runInterval:  {{ .Values.check.deployment.runInterval }}
+  timeout: {{ .Values.check.deployment.timeout }}
   podSpec:
     containers:
     - name: deployment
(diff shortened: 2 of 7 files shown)


deploy/helm/kuberhealthy/templates/check-reaper.yaml
```diff
@@ -1,3 +1,5 @@
+{{- if .Values.checkReaper.enabled }}
```
joshulyne, line 1: I'm hesitant to make check-reaper configurable in helm... Only concern being that if folks are using Kuberhealthy 2.1.0+, checks will not get cleaned up. If we were to make it configurable in Helm, it would be definitely have to be enabled by default and customers should be warned about disabling it

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -90,13 +98,16 @@ check:
...
+    runInterval: 5m
+    timeout: 15m
     image:
       repository: kuberhealthy/pod-status-check
       tag: v1.2.2
     extraEnvs:
     nodeSelector: {}
 
 checkReaper:
+  enabled: true
```
joshulyne, line 110: Mentioned already in the above comment, but a warning would be nice here if this is disabled! Also wondering if this should be contingent on what Kuberhealthy version they're using. I say that if they're using Kuberhealthy 2.1.0+, this should be enabled!

joshulyne:
thanks for addressing your own issue :) added a comment about the `checkReaper` but other than that, lgtm! 

@jonnydawg @chadbitzer, i know you guys did most of the helm configurations, so would be nice to get your reviews!

NissesSenap:
No worries.

I added a extra varning in the values file as you pointed out @joshulyne.
Hope you think that is enough.

deploy/helm/kuberhealthy/templates/khcheck-deployment.yaml
```diff
@@ -6,16 +6,16 @@ metadata:
...
+  timeout: {{ .Values.check.deployment.timeout }}
   podSpec:
     containers:
     - name: deployment
       image: {{ .Values.check.deployment.image.repository }}:{{ .Values.check.deployment.image.tag }}
       imagePullPolicy: IfNotPresent
       env:
         - name: CHECK_TIME_LIMIT
-          value: *deployment_check_timeout
+          value: {{ .Values.check.deployment.timeout }}
```
jonnydawg, line 18: Thanks! I was wondering if there was a better way to do this (yaml anchors and nodes) -- it seems like helm is a perfect solution!

Issue #361, opened 2020-03-16
service account missing in deployment-check


NissesSenap:
**Describe the feature you would like and why you want it**
I need the possibility to add a service account to the deployment that that deployment-check creates through the khchecks.comcast.github.io deployment or something similar.

**Additional context**
I'm using openshift and due to the extra security context i need to be able to set a service account that got the possibility to create containers that runs as root user. The pod is currently started with the default serviceaccount and if i patch it, it get's replaced by the operator.

jonnydawg:
Thanks for the work! Your PR was merged :)

PR #425, opened 2020-04-18
Adding CHECK_SERVICE_ACCOUNT to khcheck deployment

NissesSenap:
Should have done this in #361 but forgot.

This way people don't have to read the code to know about this setting.
The default SA is the default account that will be used by default so this won't matter for anyone who doesn't want to overwrite it like me..

### DIFF
--- a/deploy/kuberhealthy-prometheus-operator.yaml
+++ b/deploy/kuberhealthy-prometheus-operator.yaml
@@ -471,6 +471,8 @@ spec:
           value: "4"
         - name: CHECK_DEPLOYMENT_ROLLING_UPDATE
           value: "true"
+        - name: CHECK_SERVICE_ACCOUNT
+          value: "default"
       resources:
         requests:
           cpu: 25m
--- a/deploy/kuberhealthy-prometheus-operator.yaml
+++ b/deploy/kuberhealthy-prometheus-operator.yaml
@@ -471,8 +471,6 @@ spec:
           value: "4"
         - name: CHECK_DEPLOYMENT_ROLLING_UPDATE
           value: "true"
-        - name: CHECK_SERVICE_ACCOUNT
-          value: "default"
       resources:
         requests:
           cpu: 25m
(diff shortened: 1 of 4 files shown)


jonnydawg:
Looks good from the testing on my end. I think it would also be good to clean up the deployment-sa that is defined in the .yaml files for the deployment check and switch to using the default sa.

deploy/kuberhealthy-prometheus-operator.yaml
```diff
@@ -471,6 +471,8 @@ spec:
           value: "4"
         - name: CHECK_DEPLOYMENT_ROLLING_UPDATE
           value: "true"
+        - name: CHECK_SERVICE_ACCOUNT
```
jonnydawg, line 474: These files under deploy are auto-generated by github actions based on the contents in the helm directory -- there is no need to touch these as they will get overwritten. Placing them in the `values.yaml` and the `khcheck-deployment.yaml` files should avoid this overwrite.

NissesSenap:
The deployment-sa account should still be needed to be able to deploy the pods.

The reason why I need the CHECK_SERVICE_ACCOUNT function in Openshift is due to that I create a SA account that I allow to run containers as root which is needed since the deploy check deploys nginx and sadly nginx always is run as root.

I could of course use deployment-sa to do this but since this SA don't need to be able to run containers that runs as root after the changes that was recently done to the Dockerfiles I don't want to give this account any extra access that isn't needed.

Or did i misunderstand you?

deploy/kuberhealthy-prometheus-operator.yaml
NissesSenap: Ahh I have seen the create-flat-files.sh but I never opened it, a nice solution.
I will push a fix that just changes the values.yaml.

jonnydawg:
I think I misunderstood you here.

Would you suggest moving away from using nginx servers?

NissesSenap:
I don't think it's worth the job to move away from nginx.
In a dream scenario Red Hat or nginx would release a non root container but I don't think they have nor will and I don't want to try to maintain something like that on my own.

The good thing with nginx is that everyone knows what it is so it's a very simple deployment and service to debug if something strange would happen.

PR #433, opened 2020-04-23
MOD:change string contain 'not found' to k8s error

calmkart:
change error in k8s api return `strings.Contains(err.Error(), "not found")` to `k8sErrors.IsNotFound(err)`
the `k8sErrors "k8s.io/apimachinery/pkg/api/errors"` can easily classify the cause of the error int k8s api.


jonnydawg:
Whoa, thanks for making this change! I did not know this was available. I will quickly do some testing on this.

PR #438, opened 2020-04-24
MOD:fix ownerRef bug && create ownerRef for khcheck-pod khstate to khcheck

calmkart:
now create check pod ownerReference by get the kuberhealthy pod `os.hostname`
https://github.com/Comcast/kuberhealthy/blob/3309e2b740e635f00bcd257c5ba0be5209ddaed2/pkg/checks/external/util/main.go#L1-L45

This will cause a bug and directly cause kuberhealhty to be unavailable in my cluster. Because the os.hostname of my kuberhealthy is not equal to the name of the pod

It is also inappropriate to point the ownerReference of the check pod to the kuberhealthy pod. The more logical process should be kuberhealthy khcheck cr control khstate and the corresponding check pod. (That is, when a khcheck cr is deleted, the corresponding khstate and check pod will be automatically deleted)

This also has the advantage that we no longer need to use the timer to query whether khcheck cr is deleted, and then delete the corresponding khstate. Now, when khcheck cr is deleted, its bound khstate and check pod will be automatically cleared.


calmkart:
for #439 
@jonnydawg  
PTAL,thanks

calmkart:
add a new commit

After testing, khstate's OwnerReference can be successfully written, but occasionally khstate cascade delete failures (check pod cascade delete works normally). This should be a bug in some versions of k8s itself, so a deleteKhstate goroutine was added to delete the corresponding khstate cr when the watch goes to the khcheck cr object deletion event

calmkart:
@integrii 
PTAL

integrii:
Thanks. I have been working though PRs after being away for a couple weeks. We have a new baby!

I am going to check this out tomorrow.

joshulyne:
Hmm this is interesting -- @NissesSenap helped us with the original ownerRef implementation in this pr: https://github.com/Comcast/kuberhealthy/pull/378. The reason why we tied the ownerReference to the Kuberhealthy pod was because the Kuberhealthy pod was the one responsible for creating the checker pods, the khcheck is a cr that requests the creation of the checker pod. This is definitely up for further debate. I am a fan of tying up all three resources, the khcheck, khstate, and the khpod -- that a deletion of the khcheck should also delete the khstate and khpod

cmd/kuberhealthy/crd.go
```diff
@@ -64,7 +65,7 @@ func sanitizeResourceName(c string) string {
 }
 
 // ensureStateResourceExists checks for the existence of the specified resource and creates it if it does not exist
-func ensureStateResourceExists(checkName string, checkNamespace string) error {
+func ensureStateResourceExists(checkName string, checkNamespace string, k *Kuberhealthy) error {
```
integrii, line 68: Could we refactor this function so that we pass in the owner ref instead of the `Kuberhealthy` instance?  This would shorten the code here and make it assume less things.  This would of course mean finding all the calls to this func and augmenting them with the proper ownerRef data.  That owner ref could probably be a global in the invoking Kuberhealthy program, though - which is faster than querying it every time.

What do you think?

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -302,6 +235,7 @@ func (k *Kuberhealthy) watchForKHCheckChanges(ctx context.Context, c chan struct
 				c <- struct{}{}
 			case watch.Deleted:
 				log.Debugln("khcheck monitor saw a deleted event")
+				go k.deleteRelatedKhstate(*khc.Object.(*khcheckcrd.KuberhealthyCheck))
```
integrii, line 238: Can you do your type assertion this on the prior line and check the error value for it while you're there?

https://golang.org/ref/spec#Type_assertions

cmd/kuberhealthy/kuberhealthyCheck.go
```diff
@@ -45,4 +47,8 @@ type KuberhealthyCheck interface {
 	// When the check completes and returns, we assume it is done shutting
 	// down.
 	Shutdown() error
+	// GenOwnerRef can generate the OwnerReference for objects(pod/cr)
```
integrii, line 50: This interface was from Kuberhealthy v1.  In v1 we had lots of check structs that would be thrown in from different packages.  In v2, we only have one!  That one is the external checker, which then runs all other checks.

I think it's okay to add this to the interface for now, but I wanted to document the intent to remove this interface.

pkg/checks/external/main.go
```diff
@@ -1265,6 +1269,28 @@ func (ext *Checker) getPodClient() typedv1.PodInterface {
...
 
+// GenOwnerRef return the OwnerReference for objects(pod/cr) create by the khcheck checker
+func (ext *Checker) GenOwnerRef() ([]metav1.OwnerReference, error) {
+	khcheck, err := ext.getCheck()
+	if err != nil {
+		return nil, err
+	}
+
+	controller := true
+	blockOwnerDeletion := true
```
integrii, line 1280: Do you have the docs for this?  What does `blockOwnerDeletion` mean here?  Sorry for being lazy :-)

cmd/kuberhealthy/crd.go
calmkart: we need Kuberhealthy instance to run 
```
checker, err := k.getCheck(checkName, checkNamespace)
if err != nil {
	return err
}
// Get ownerReference for the kuberhealthy khstate
ownerRef, err := checker.GenOwnerRef()
if err != nil {
	return errors.New("Failed to getOwnerReference for khstate: " + initialState.Name + ", err: " + err.Error())
}
```
but i think make `ensureStateResourceExists ` to a method for Kuberhealthy instance is a good idea.
```
func (k *Kuberhealthy) ensureStateResourceExists(checkName string, checkNamespace string) error {
.......
}
```
i'll fix it

cmd/kuberhealthy/kuberhealthy.go
calmkart: watch event
```
// Event represents a single event to a watched resource.
// +k8s:deepcopy-gen=true
type Event struct {
	Type EventType

	// Object is:
	//  * If Type is Added or Modified: the new state of the object.
	//  * If Type is Deleted: the state of the object immediately before deletion.
	//  * If Type is Bookmark: the object (instance of a type being watched) where
	//    only ResourceVersion field is set. On successful restart of watch from a
	//    bookmark resourceVersion, client is guaranteed to not get repeat event
	//    nor miss any events.
	//  * If Type is Error: *api.Status is recommended; other types may make sense
	//    depending on context.
	Object runtime.Object
}
```
It seems that only type is Error will return * api.Status, otherwise it will return the watch object itself

But adding a type check is also good, I will fix it

pkg/checks/external/main.go
calmkart: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

blockOwnerDeletion is set to true, when we use the Foreground delete mode, the delete will be blocked. Although we use the Background delete mode, it does not exclude users from manually deleting the specified delete mode. Setting it to true here is relatively safe, and there will be no major problems.

joshulyne:
> Hmm this is interesting -- @NissesSenap helped us with the original ownerRef implementation in this pr: #378. The reason why we tied the ownerReference to the Kuberhealthy pod was because the Kuberhealthy pod was the one responsible for creating the checker pods, the khcheck is a cr that requests the creation of the checker pod. This is definitely up for further debate. I am a fan of tying up all three resources, the khcheck, khstate, and the khpod -- that a deletion of the khcheck should also delete the khstate and khpod

@integrii ^^

pkg/checks/external/main.go
integrii: Thank you!

cmd/kuberhealthy/kuberhealthy.go
integrii: Thank you!

cmd/kuberhealthy/crd.go
integrii: I think a method works here. Thank you.

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -324,6 +263,26 @@ func (k *Kuberhealthy) watchForKHCheckChanges(ctx context.Context, c chan struct
...
 }
 
+// delete the khstate object
+func (k *Kuberhealthy) deleteRelatedKhstate(khCheck khcheckcrd.KuberhealthyCheck) {
+	khState, err := khStateClient.Get(metav1.GetOptions{}, stateCRDResource, khCheck.GetName(), khCheck.GetNamespace())
+	if err != nil {
+		if k8sError.IsNotFound(err) {
+			return
+		}
+		log.Errorln(fmt.Errorf("khState reaper: error when get the invalid khstate: %w", err))
```
integrii, line 273:
```suggestion
		log.Errorln(fmt.Errorf("khState reaper: error when fetching khstate that needs cleaned up: %w", err))
```

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -324,6 +263,26 @@ func (k *Kuberhealthy) watchForKHCheckChanges(ctx context.Context, c chan struct
...
+	}
+
+	_, err = khStateClient.Delete(khState, stateCRDResource, khState.GetName(), khState.GetNamespace())
+	if err != nil {
+		if k8sError.IsNotFound(err) {
+			return
+		}
+		log.Errorln(fmt.Errorf("khState reaper: error when removing invalid khstate: %w", err))
```
integrii, line 281:
```suggestion
		log.Errorln(fmt.Errorf("khState reaper: error when removing expired khstate: %w", err))
```

pkg/checks/external/main.go
```diff
@@ -1265,6 +1269,28 @@ func (ext *Checker) getPodClient() typedv1.PodInterface {
...
+		metav1.OwnerReference{
+			APIVersion:         CRDGroup + "/" + CRDVersion,
+			Kind:               "KuberhealthyCheck",
+			Name:               khcheck.Name,
+			UID:                khcheck.UID,
+			Controller:         &controller,
+			BlockOwnerDeletion: &blockOwnerDeletion,
+		},
+	}, nil
+}
```
integrii, line 1292: I would prefer to have the owner reference target the master Kuberhealthy instance.  This way, if the master pod crashes or gets removed, the checker pods it was managing get cleaned up.

If the CRD is the owner, then the checker pods will never really get cleaned up until the user deletes the check configuration.  The `khcheck` is like a user configuration and the master pod is like the owner, I think.

That said, I think it would make sense to possibly make the `khstate` objects owned by the `khcheck` object that pair to... If you feel up to splitting that out, I think it would be a nice improvement.  If not, we can make an issue to track the idea and flip this over to the master pod being master.

calmkart: If we point the ownerreference of the check pod to the kuberhealth master pod, then when we delete the kuberhealth master pod, the check pod will indeed be cleaned, but the khchecks CR will not be cleared. In any case, after all, only after the creation of khcheck CR can the check pod be created. If you want to get a complete cleanup, I think the relationship chain should be such a kuberhealth master Pod - > khcheck Cr - > khstate Cr - > check pod. I think we should add a logic to point the ownerreference of khcheck CR to kuberhealth master pod. What do you think? If possible, I will add this logic after the holiday.

cmd/kuberhealthy/kuberhealthy.go
calmkart: done

cmd/kuberhealthy/kuberhealthy.go
calmkart: done

pkg/checks/external/main.go
NissesSenap: So this is an issue that all operators have, we want to make sure that everything is as clean as possible no matter what the end user does.

But the standard is the CR (Custom Resource) is the owner of the rest of the things that gets created and the controller, aka the operator is just sitting and listening for stuff to do.

@integrii as you might remember my first stab at ownerRef issue was that I wanted to set the khcheck as the ownerRef even if I did a really bad job at it :). And since i didn't have the technical knowledge I agreed to do it the current way since it was better then nothing from my point of view.

Sure if we remove the operator it self we will still have khchecks and khstats left but those are just entries in etcd and it won't take any resources and all of the check pods will go to complete or error.

But on the other hand if we bind the ownerRef to the kuberhealthy master pod sure all of the pods will disappear but this might happen just because the kuberhealthy master pod gets migrated to another node due to the physical node crashes or something else and that will ofc break all the checks and we will get bad data in our monitoring.
And when I think about it, this actually removes the functionality of the secondary operator, since if the master operator ever dies the secondary will never take over with a current check, instead all of the checks will be restarted. Fine the secondary node will take over faster but it's only with a few seconds.

But i definitely agree with you that the khstate should be owned by khcheck.

So in short let khcheck be the owner of the check pod and khstat

 khstat
      |^
khcheck -> checker pod

calmkart: @integrii  I cannot set the ownerref of khcheck in the code because khcheck is created by the user and not by kuberhealthy. I still think it is more reasonable to point the ownerRef of the check pod to khcheck. In other words, it may not be the final state, but it is more convenient to use than the current way.

integrii:
I definitely don't want `KuberhealthyCheck` to be the owner of checker pods... Master instances die or get shut down much more than KuberhealthyCheck configurations get removed.  If oyu could, please do two things:

- Wait on purging `KuberhealthyState` resources.  We can make a new issue to tackle this.
- Change checker pod ownership to be the current pod starting the check (the master Kuberhealthy instance).  I think this is how we do it now after changes from @NissesSenap.

Also, @NissesSenap you should know that any failover event will restart all checks as the secondary Kuberhealthy instance takes over... Just like when any `KuberhealthyCheck` changes are made.

Let me know what you think @calmkart

calmkart:
> I definitely don't want `KuberhealthyCheck` to be the owner of checker pods... Master instances die or get shut down much more than KuberhealthyCheck configurations get removed. If oyu could, please do two things:
> 
> * Wait on purging `KuberhealthyState` resources.  We can make a new issue to tackle this.
> * Change checker pod ownership to be the current pod starting the check (the master Kuberhealthy instance).  I think this is how we do it now after changes from @NissesSenap.
> 
> Also, @NissesSenap you should know that any failover event will restart all checks as the secondary Kuberhealthy instance takes over... Just like when any `KuberhealthyCheck` changes are made.
> 
> Let me know what you think @calmkart

Okay, I understand.
Thank you very much for your patient reply.
I'll close this pr

PR #443, opened 2020-04-27
external checks: add new network connection check

bavarianbidi:
network-connection-check allows to test if a connection to a network target
could be done (via tcp or udp) and a failed connection is expected.


cmd/network-connection-check/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+The check runs every 2 minutes (spec.runInterval), with a check timeout set to 15 minutes (spec.timeout). If the check
+does not complete within the given timeout and target should be reachable it will report a timeout error on the status page. The check takes in a
+`CHECK_POD_TIMEOUT` environment variable that ensures the the pod runs the connection lookup within the timeout.
+If the check does not complete within the given timeout and target should not be reachable it will report a success on the status page.
+
+To verify connections, apply another KHCheck configuration file with a different `CONNECTION_TARGET` environment variable.
+`CONNECTION_TARGET` accepts `tcp://` or `udp://` (e.g. `udp://10.16.12.10:9000` or `tcp://10.16.12.11:8080`)
+
+#### DNS Status Check Kube Spec:
```
NissesSenap, line 22: Should probably say network-connection-check

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -106,6 +106,16 @@ check:
       tag: v1.2.2
     extraEnvs:
     nodeSelector: {}
+  networkConnection:
```
integrii, line 109: We don't have to add every new check to the helm chart here - we can simply add them to the check registry.

Actually, I don't even mind if the check is sourced from another git repository and image host entirely!  This means you could self-publish this check and then open a small PR to update the `EXTERNAL_CHECKS_REGISTRY.md` file.

deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
```diff
@@ -0,0 +1,32 @@
...
+---
+apiVersion: comcast.github.io/v1
+kind: KuberhealthyCheck
+metadata:
+  name: network-connection-check
+  namespace: {{ .Release.Namespace }}
+spec:
+  runInterval:  {{ .Values.check.networkConnection.runInterval }}
+  timeout: {{ .Values.check.networkConnection.timeout }}
+  podSpec:
```
integrii, line 11:
```suggestion
  podSpec:
      {{- if .Values.securityContext.enabled }}
    securityContext:
      runAsUser: {{ .Values.securityContext.runAsUser }}
      fsGroup: {{ .Values.securityContext.fsGroup }}
    {{- end}}
```

deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
```diff
@@ -0,0 +1,32 @@
...
+      env:
+        - name: CHECK_TIME_LIMIT
+          value: {{ .Values.check.networkConnection.timeout }}
+{{- range $key, $value := .Values.check.networkConnection.extraEnvs }}
+        - name: {{ $key }}
+          value: {{ $value | quote }}
+{{- end }}
+      resources:
+        requests:
+          memory: 5Mi
```
integrii, line 25:
```suggestion
          memory: 5Mi
        {{- if .Values.securityContext.enabled }}
        securityContext:
          allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation }}
          readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
        {{- end}}
```

cmd/network-connection-check/main.go
```diff
@@ -0,0 +1,187 @@
...
+type Checker struct {
+	client            *kubernetes.Clientset
+	connectionTarget  string
+	targetUnreachable bool
+}
+
+func init() {
+
+	// Grab and verify environment variables and set them as global vars
+	CheckTimeout := os.Getenv("CHECK_POD_TIMEOUT")
```
jonnydawg, line 50: I think checker pods are able to calculate how much time they have for a check run now with the `KH_CHECK_RUN_DEADLINE` environment variable which will give you a more accurate check deadline. This is still appropriate though.

integrii: #450

bavarianbidi: wasn't aware of this ENV ...
... have to rework this PR.
I would update the documentation regarding all injected ENV vars in this PR.

jonnydawg:
Some deadline documentation was touched in #452 and #454

integrii:
This one is currently waiting for the feedback about the check pod timeout ENV var to be integrated before merging.

deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
bavarianbidi: from my point of view we don't need the `securityContext` here.
* the application itself doesn't need any extended permissions (only does a tcp/udp connection check, no additional file operations).
* container is build with a numeric user and point to `nobody` (https://github.com/Comcast/kuberhealthy/pull/443/files#diff-b80ba549f0d3f00bfc542d54b0e03f7cR10) 
* if a `psp` is deployed where the usage of `root` is forbidden, kubernetes can schedule pods without `securityContext` when the `USER` field is a numeric value.

deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
bavarianbidi: from my point of view we don't need the `securityContext` here.
* the application itself doesn't need any extended permissions (only does a tcp/udp connection check, no additional file operations).
* container is build with a numeric user and point to `nobody` (https://github.com/Comcast/kuberhealthy/pull/443/files#diff-b80ba549f0d3f00bfc542d54b0e03f7cR10) 
* if a `psp` is deployed where the usage of `root` is forbidden, kubernetes can schedule pods without `securityContext` when the `USER` field is a numeric value.

bavarianbidi:
> This one is currently waiting for the feedback about the check pod timeout ENV var to be integrated before merging.

was AFK the last days, will test tomorrow morning and give feedback here

bavarianbidi:
Will Switch back to the previous implementation with a separate `TIMEOUT` variable...
... the current version sometimes fail if connection target is expected to be unreachable. time until the check pod will start does take a while ( image has to be pulled, ... ) and the current hard coded `-5s` isn't flexible for this and a separate parameter make sense (IMHO)

bavarianbidi:
@jonnydawg / @integrii 

could you please take a look again ;-)

deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
integrii: Gotcha.  As long as it works in these kind of locked down environments, I don't have a strong opinion.

PR #448, opened 2020-04-29
corrected Dockerfile back to use COPY instead of ADD

jdowni000:
Corrected Dockerfile's to use COPY instead of ADD to pass Dockerfile check in github actions


jonnydawg:
Are you going to tackle the main Dockerfile too?

jdowni000:
> Are you going to tackle the main Dockerfile too?

I

Issue #441, opened 2020-04-24
pod-status helm missing SA


NissesSenap:
**Describe the bug**
The helm chart is currently missing any definiton of an SA for pod-status
Due to this I get the following error when running the pod-status check.

2020-04-23T07:13:50.5539520Z - apiVersion: comcast.github.io/v1
2020-04-23T07:13:50.5539633Z   kind: KuberhealthyState
2020-04-23T07:13:50.5539717Z   metadata:
2020-04-23T07:13:50.5539978Z     creationTimestamp: "2020-04-23T07:12:50Z"
2020-04-23T07:13:50.5540091Z     generation: 7
2020-04-23T07:13:50.5540303Z     name: pod-status
2020-04-23T07:13:50.5540406Z     namespace: kuberhealthy
2020-04-23T07:13:50.5540523Z     resourceVersion: "1221"
2020-04-23T07:13:50.5540840Z     selfLink: /apis/comcast.github.io/v1/namespaces/kuberhealthy/khstates/pod-status
2020-04-23T07:13:50.5541115Z     uid: 2c2622b6-b979-409f-bb48-136bad0f3162
2020-04-23T07:13:50.5541200Z   spec:
2020-04-23T07:13:50.5541462Z     AuthoritativePod: kuberhealthy-5c9996bfdc-ncpxc
2020-04-23T07:13:50.5541567Z     Errors:
2020-04-23T07:13:50.5541848Z     - **'pods is forbidden: User "system:serviceaccount:kuberhealthy:default" cannot
2020-04-23T07:13:50.5542169Z       list resource "pods" in API group "" in the namespace "kuberhealthy"'**
2020-04-23T07:13:50.5542582Z     LastRun: "2020-04-23T07:13:39.269593722Z"
2020-04-23T07:13:50.5542705Z     Namespace: kuberhealthy
2020-04-23T07:13:50.5542809Z     OK: false
2020-04-23T07:13:50.5542925Z     RunDuration: 19.019638943s
2020-04-23T07:13:50.5543171Z     uuid: b979eb4a-3769-4c84-8591-ed49342a81f6

**Steps To Reproduce**
Currently working on creating e2e tests for kuberhealthy.
These logs can be viewed in:
https://github.com/NissesSenap/kuberhealthy/runs/611166285?check_suite_focus=true

In the commit you can also see exactly how the environment looks like.

Clon my fork and use the actions branch.
https://github.com/NissesSenap/kuberhealthy/tree/actions

[...]

PR #456, opened 2020-05-01
Iss 441 pod sa

2infinitee:
I think my PR includes a lot of changes so please let me know if I over did it or likewise need to apply it other areas that I may of left out. I basically added the service accounts to the pod-status and pod-restart checks. I also updated the `deployment.yaml` in the cmd/pod-status-check and cmd/pod-restart-check paths as well.

I also noticed that I did not add security context to the `deployment.yaml` files in the cmd paths so I updated those too. I also updated the kuberhealthy.yaml file in the deploy file path as well.

### DIFF
--- a/deploy/helm/kuberhealthy/templates/khcheck-pod-restarts.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-pod-restarts.yaml
@@ -42,4 +42,41 @@ spec:
 {{- toYaml .Values.check.podRestarts.nodeSelector | nindent 6 }}
     {{- end }}
     restartPolicy: Never
+    serviceAccountName: pod-restart-sa
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: pod-restart-rb
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: pod-restart-role
+subjects:
+  - kind: ServiceAccount
+    name: pod-restart-sa
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: pod-restart-role
+  namespace: {{ .Release.Namespace }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - watch
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pod-restart-sa
+  namespace: {{ .Release.Namespace }}
 {{- end }}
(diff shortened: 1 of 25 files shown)


2infinitee:
https://github.com/Comcast/kuberhealthy/issues/441

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -93,7 +93,7 @@ check:
       HOSTNAME: "kubernetes.default"
     nodeSelector: {}
   podRestarts:
-    enabled: false
```
jonnydawg, line 96: I think you should switch these back to false -- I'm not sure if these checks are considered `out-of-the-box` yet

jonnydawg:
Have you tested the new checks with these security context changes?

deploy/helm/kuberhealthy/templates/khcheck-pod-restarts.yaml
```diff
@@ -42,4 +42,41 @@ spec:
 {{- toYaml .Values.check.podRestarts.nodeSelector | nindent 6 }}
     {{- end }}
     restartPolicy: Never
-{{- end }}
+    serviceAccountName: pod-restart-sa
```
integrii, line 45: This is fine here, but in the future you don't need to label service accounts with `-sa` at the end.  It will be clear when you fetch objects of type `serviceAccount`.

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -103,7 +103,7 @@ check:
       MAX_FAILURES_ALLOWED: "10"
     nodeSelector: {}
   podStatus:
-    enabled: false
+    enabled: true
```
integrii, line 106:
```suggestion
    enabled: false
```

I agree with @jonnydawg - let's be really careful about what default functionality we ship.  We should also be careful when adding additional checks by default because users updating may be surprised by the additional checks.

deploy/helm/kuberhealthy/templates/khcheck-pod-restarts.yaml
2infinitee: Noted, I didn't know if we wanted to stick to a convention or something.

deploy/helm/kuberhealthy/values.yaml
2infinitee: I think you're right I might of included them because I was generating the new templates.

2infinitee:
> Have you tested the new checks with these security context changes?

Yes I have, they were tested when I made those changes and now with these changes as well.

deploy/helm/kuberhealthy/values.yaml
2infinitee: Thanks, will be more careful next time.

2infinitee:
@integrii same with this pr - merged in master and checks have passed, feel free to merge.

deploy/helm/kuberhealthy/templates/khcheck-pod-status.yaml
```diff
@@ -37,8 +37,45 @@ spec:
...
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pod-status-sa
```
integrii, line 79: No big deal here, but in Kubernetes things are already typed, so you don't need to include `-sa` on the names.

integrii:
I am sad to see the extra complex specs added in, but I suppose this is probably important for OpenShift users and clusters that don't allow a default serviceAccount any permissions at all.

PR #457, opened 2020-05-01
changing match uuid logic try duration from 1 to 3 secs

2infinitee:
increasing duration of sleep for each try from 1 to 3 secs before loop errors out.

### DIFF
--- a/pkg/checks/external/main.go
+++ b/pkg/checks/external/main.go
@@ -350,12 +350,12 @@ func (ext *Checker) setUUID(uuid string) error {
 	// This for-loop is to verify the pod uuid is properly set with the api server before the checker pod is started.
 	tries := 0
 	for {
-		if tries >= 3 {
+		if tries >= 9 {
 			return fmt.Errorf("failed to verify uuid match %s after %d tries", checkState.Spec.CurrentUUID, tries)
 		}
 		tries++
-		ext.log("Waiting 3 second before checking " + ext.Name() + " uuid.")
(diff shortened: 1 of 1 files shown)


pkg/checks/external/main.go
```diff
@@ -354,8 +354,8 @@ func (ext *Checker) setUUID(uuid string) error {
 			return fmt.Errorf("failed to verify uuid match %s after %d tries", checkState.Spec.CurrentUUID, tries)
 		}
 		tries++
-		ext.log("Waiting 1 second before checking " + ext.Name() + " uuid.")
-		time.Sleep(time.Second)
+		ext.log("Waiting 3 second before checking " + ext.Name() + " uuid.")
+		time.Sleep(time.Second * 3)
```
jonnydawg, line 358: I think we should up the limit of retries instead of upping the time in-between retries.

integrii: Agree with @jonnydawg .  Could you change this on the `tries` variable instead please?  That way we don't add so much extra delay here...

Maybe set `tries` to a maximum of 9 and drop the sleep back down to 1s.

2infinitee:
@integrii merged master into this branch and it seemed to resolved all checks. Feel free to merge.

pkg/checks/external/main.go
2infinitee: done, changed from 3 secs to 1 and also up the number of tries to 9

PR #461, opened 2020-05-04
kuberhealthy: make kuberhealthy namespace aware

bavarianbidi:
in a shared cluster it's not possible to have multiple independent
kuberhealthy deployments (crd, deployment, kuberhealthy checks, ...).

current main reason is, that kuberhealthy controller fetch all
kuberhealthychecks from all namespaces even though all functions
currently accept a *namespace* argument.

within this PR you can opt in kuberhealthy with the command line
argument --listenNamespace and kuberhealthy only act in the defined
namespace. If --listenNamespace isn't defined, behavior is like before
and all khchecks in all namespaces are fetched.

# known issues
- [x] set UUID back seems to be broken --> fixed/solved after rebase on master

### DIFF
--- a/cmd/kuberhealthy/main.go
+++ b/cmd/kuberhealthy/main.go
@@ -34,13 +34,14 @@ import (
 )
 
 // status represents the current Kuberhealthy OK:Error state
-var kubeConfigFile = filepath.Join(os.Getenv("HOME"), ".kube", "config")
+var kubeConfigFile = filepath.Join(os.Getenv("KUBECONFIG"))
 var listenAddress = ":8080"
 var podCheckNamespaces = "kube-system"
 var podNamespace = os.Getenv("POD_NAMESPACE")
 var isMaster bool                  // indicates this instance is the master and should be running checks
 var upcomingMasterState bool       // the upcoming master state on next interval
 var lastMasterChangeTime time.Time // indicates the last time a master change was seen
+var listenNamespace string         // namespace to listen (watch/get) on
 
 var terminationGracePeriod = time.Minute * 5 // keep calibrated with kubernetes terminationGracePeriodSeconds
 
@@ -104,6 +105,7 @@ func init() {
 	flaggy.Bool(&enableForceMaster, "", "forceMaster", "Set to true to enable local testing, forced master mode.")
 	flaggy.Bool(&enableDebug, "d", "debug", "Set to true to enable debug.")
 	flaggy.String(&logLevel, "", "log-level", fmt.Sprintf("Log level to be used one of [%s].", getAllLogLevel()))
+	flaggy.String(&listenNamespace, "", "listenNamespace", "The namespace kuberhealthy is responsible for.")
 	// Influx flags
 	flaggy.String(&influxUsername, "", "influxUser", "Username for the InfluxDB instance")
 	flaggy.String(&influxPassword, "", "influxPassword", "Password for the InfluxDB instance")
@@ -241,14 +243,14 @@ func initKubernetesClients() error {
 	kubernetesClient = kc
 
 	// make a new crd check client
-	checkClient, err := khcheckcrd.Client(checkCRDGroup, checkCRDVersion, kubeConfigFile, "")
+	checkClient, err := khcheckcrd.Client(checkCRDGroup, checkCRDVersion, kubeConfigFile, listenNamespace)
 	if err != nil {
 		return err
 	}
 	khCheckClient = checkClient
 
 	// make a new crd state client
-	stateClient, err := khstatecrd.Client(stateCRDGroup, stateCRDVersion, kubeConfigFile, "")
+	stateClient, err := khstatecrd.Client(stateCRDGroup, stateCRDVersion, kubeConfigFile, listenNamespace)
 	if err != nil {
 		return err
 	}
--- a/cmd/kuberhealthy/main.go
+++ b/cmd/kuberhealthy/main.go
@@ -34,7 +34,7 @@ import (
 )
 
 // status represents the current Kuberhealthy OK:Error state
-var kubeConfigFile = filepath.Join(os.Getenv("KUBECONFIG"))
+var kubeConfigFile = filepath.Join(os.Getenv("HOME"), ".kube", "config")
 var listenAddress = ":8080"
 var podCheckNamespaces = "kube-system"
 var podNamespace = os.Getenv("POD_NAMESPACE")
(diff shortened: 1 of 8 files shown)


bavarianbidi:
As an alternative ( or additional flag ) we could think about an additional label on the kuberhealthy checks ( similar to ingress class of traefik - where you can define on the ingress object, which ingress controller should handle the ingress definition )

jonnydawg:
If I understand correctly, you are running into issues where your customers could also implement their own kuberhealthy with checks and those are conflicting with your checks?

The external check status endpoint can take namespace parameters that will filter out checks for specified namespaces `?namespace=kuberhealthy,kube-system` -- not sure if this helps or not.

I do think a --listenNamespace flag could do the trick here, but I think the original intent was to have one operator be running from the kuberhealthy namespace.

pkg/khstatecrd/functions.go
```diff
@@ -97,19 +137,42 @@ func (c *KuberhealthyStateClient) List(opts metav1.ListOptions, resource string,
...
 		Watch()
+
+	if len(namespace) > 0 {
+		clientResponse, err = c.restClient.Get().
+			Resource(resource).
+			NamespaceIfScoped(namespace, true).
+			VersionedParams(&opts, scheme.ParameterCodec).
+			Timeout(timeout).
+			Watch()
+	}
```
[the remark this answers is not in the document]
bavarianbidi: sorry, forget this ;-) just rebased and removed temp. helper code

bavarianbidi:
> If I understand correctly, you are running into issues where your customers could also implement their own kuberhealthy with checks and those are conflicting with your checks?
> 
> The external check status endpoint can take namespace parameters that will filter out checks for specified namespaces `?namespace=kuberhealthy,kube-system` -- not sure if this helps or not.
> 
> I do think a --listenNamespace flag could do the trick here, but I think the original intent was to have one operator be running from the kuberhealthy namespace.

yes, you are right, broken `kuberhealthy-checks` could conflict with our checks. I was aware of the `namespace-parameter` in the status endpoint, but this could lead to a lot of log messages in kuberhealthy and i would keep it more clean.

bavarianbidi:
> Thanks for contributing! Looks like you did a good bit of discovery in the codebase and your own testing. That really helps.
> 
> I think you may want to be careful that your customers' Kuberhealthy instances don't have permission to fetch your `khcheck` resources. Kuberhealthy was designed to be run once per cluster (as a global operator). I think your contributed flag could be useful, so I am willing to merge the PR once reviewed sufficiently, but it may not be the best way to segregate your users alerts and metrics...
> 
> Would it be possible to filter alerts from metrics on the AlertManager rule somehow?
> 
> Could you explain what you are seeing in the known issue `set UUID back seems to be broken`?

fetching `khcheck` resources isn't the issue here (because there aren't any secrets there).
Sure, kuberhealthy works well to run once per cluster but the CRD itself are namespace scoped and as you can see in this PR, there aren't much code-changes necessary to add the `namespace` because it was mostly there ;-)

It is absolute possible to filter out customer defined `khchecks` in our alerting system. The initial integration was done with that but we have found that this deployment mode could conflict with customers if the wan't to define `khchecks` by their own. And it make us more independent to maintain our `kuberhealthy` instances (without customer impact)

> Could you explain what you are seeing in the known issue `set UUID back seems to be broken`?
Was solved after rebase on master this morning

bavarianbidi:
just for your Info:
kuberhealthy with commit [#24587dd9](https://github.com/Comcast/kuberhealthy/pull/461/commits/24587dd93ac1f1ab4d9c50c49af82a0a0babd684) now runs since 6 hours in a dev cluster in two different namespaces (two independent kuberhealthy instances with 6 `khchecks` in each namespace) without any known issues

bavarianbidi:
@integrii 

could you please take a look ;-)

cmd/kuberhealthy/main.go
```diff
@@ -34,13 +34,14 @@ import (
 )
 
 // status represents the current Kuberhealthy OK:Error state
-var kubeConfigFile = filepath.Join(os.Getenv("HOME"), ".kube", "config")
+var kubeConfigFile = filepath.Join(os.Getenv("KUBECONFIG"))
```
integrii, line 37: Can we leave this line as:
`var kubeConfigFile = filepath.Join(os.Getenv("HOME"), ".kube", "config")`

and then add an override if the `KUBECONFIG` environment variable is set?   This way users who do not have a `KUBECONFIG` variable set will not be broken and users who do use it will see it take effect.

pkg/khstatecrd/functions.go
```diff
@@ -101,15 +100,19 @@ func (c *KuberhealthyStateClient) List(opts metav1.ListOptions, resource string,
...
-		Resource("khstates").
+
+	clientResponse, err := c.restClient.Get().
+		Resource(resource).
+		Namespace(namespace).
 		VersionedParams(&opts, scheme.ParameterCodec).
 		Timeout(timeout).
 		Watch()
+
+	return clientResponse, err
```
integrii, line 117: You can change this to simply return:

```suggestion
	return c.restClient.Get().
		Resource(resource).
		Namespace(namespace).
		VersionedParams(&opts, scheme.ParameterCodec).
		Timeout(timeout).
		Watch()
```

bavarianbidi:
> I suggested a couple little changes here, but nothing major.
> 
> The problem I see with this change is that this limits the running Kuberhealthy instance, but does not prevent other Kuberhealthy instances in the cluster from reading CRDs that aren't meant for it (and duplicating the checks).
> 
> Once one Kuberhealthy instance uses the `listenNamespace` flag, then all other instances must know to use it. Moreover, Kuberhealthy can only target one namespace when using `listenNamespace`. You can't specify multiple namespaces and adding that functionality would be significantly more code.
> 
> I think it's fine to merge for now to fit your use cases, but we should keep in mind that we've created these tricky situations...
> 
> @jdowni000 when this is merged to master, you will need to pull in master and resolve the conflcits this has created.
> 
> Justin is working on moving all Kuberhealthy startup flags to a configuration file.

i agree with you.

One possible imlementation could look like (IMHO):
1. new configuration flag (param/config file) like `kuberhealthyClass` (similar to [traefiks ingress class](https://docs.traefik.io/providers/kubernetes-ingress/#ingressclass) where you can define, which `traefik` instance should handle which `ingress` object)
1. set flag lead to
    ```
   metav1.ListOptions {
       LabelSelector: "kuberhealthyClass=<valueFromAboveArg>",
    }
    ```
   In most cases, we use `metav1.ListOptions{}` to watch/listen/get objects ([list all khstates in the cluster](https://github.com/Comcast/kuberhealthy/blob/master/cmd/kuberhealthy/kuberhealthy.go#L229))

Then it's possible to add the `kuberhealthyClass` label to `khcheck` objects. `listenNamespace` and `kuberhealthyClass` could coexist and both implementations could work side by side.

bavarianbidi:
@integrii 

thanks for your feedback!

integrii:
Good thoughts on clustering.  Let's try to think about what the world of multiple KH deployments in a single cluster looks like later - and only if we really have to.  I think the design change to support multiple instances per cluster may turn out to case a lot more chaos than we see on the surface.

PR Merged!  Thank you for sticking with this for so long.  We should be able to get this out with the 2.2 release.

PR #462, opened 2020-05-04
Update deployment check time limit

jonnydawg:
subtract 5s from time-limit;bump version

Subtracts 5s from the time limit to ensure that the check reports its last stage before timing out.
Bumps the version up a patch number.

### DIFF
--- a/cmd/deployment-check/input.go
+++ b/cmd/deployment-check/input.go
@@ -171,19 +171,11 @@ func parseInputValues() {
 	// Set check time limit to default
 	checkTimeLimit = defaultCheckTimeLimit
 	// Get the deadline time in unix from the env var
-	unixDeadline, err := kh.GetDeadline()
+	timeDeadline, err := kh.GetDeadline()
 	if err != nil {
 		log.Infoln("There was an issue getting the check deadline:", err.Error())
 	}
-	// Calculate check run duration from the deadline and now
-	deadline, err := strconv.Atoi(unixDeadline)
-	if err != nil {
-		log.Fatalln("Failed to parse unix deadline from environment.")
-	}
-	if deadline > 0 {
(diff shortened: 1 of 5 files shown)


cmd/deployment-check/input.go
```diff
@@ -182,8 +182,9 @@ func parseInputValues() {
 	}
 	if deadline > 0 {
 		log.Infoln("Parsed check deadline time from the environment:", deadline)
-		checkTimeLimit = time.Duration((int64(deadline) - time.Now().Unix()) * 1e9) // Multiply by 1,000,000,000 because that's how many nanoseconds are in a second
+		checkTimeLimit = time.Duration((int64(deadline) - time.Now().Unix() - 5) * 1e9) // Multiply by 1,000,000,000 because that's how many nanoseconds are in a second
```
integrii, line 185: Can you refactor this to parse the deadline into a `time.Time`, then use `time.After()` to determine if the time has passed?

You can also use `time.Sub()` to remove a duration (like `time.Second * 5`) from a `time.Time`.

cmd/test-external-check/main.go
```diff
@@ -46,19 +46,12 @@ func init() {
...
-	deadline, err := strconv.Atoi(unixDeadline)
-	if err != nil {
-		log.Fatalln("Failed to parse unix deadline from environment.")
-	}
-	if deadline > 0 {
-		log.Println("Parsed check deadline time from the environment:", deadline)
-		timeLimit = time.Duration((int64(deadline) - time.Now().Unix()) * 1e9) // Multiply by 1,000,000,000 because that's how many nanoseconds are in a second
-	}
+	timeLimit = timeDeadline.Sub(time.Now().Add(time.Second * 5))
+	log.Println("Check time limit set to:", timeLimit)
```
integrii, line 54: Much better!

PR #471, opened 2020-05-13
added info about errors in Prometheus' metrics

simonemex:
I'm using Prometheus and Alertmanager to send Kuberhealthy errors to a Slack channel.

With this PR I have included the errors which are causing the check failure, so they can be included in any notifications from Alertmanager.

The format of string representing errors is `error1|error2`


integrii:
This looks great!  Thank you for the contribution.  We should release this in 2.2.0.

Issue #472, opened 2020-05-13
KH_CHECK_RUN_DEADLINE environment variable required in latest test-external-check, but not defined in KH 2.1.0. 


zjhans:
**Describe the bug**
Running kuberhealthy/test-external-check:latest crashes test pod. Version requires environment variable KH_CHECK_RUN_DEADLINE which is undefined currently. 

Additionally, error message in logs needs formatting fix, spaces are missing around the variable name in the error.

Logs:
2020/05/12 17:27:43 checkClient: ERROR: kuberhealthy check deadline from environment variableKH_CHECK_RUN_DEADLINEwas blank

**Steps To Reproduce**
- Run the KH check using image _kuberhealthy/test-external-check:latest_ instead of _v1.2.1_ or earlier.
- Check failed pods log for above error

**Expected behavior**
All published versions of the test check should work if properly configured.

**Versions**
 - Cluster OS: Docker Desktop
 - Kubernetes Version: 1.16.6
 - Kuberhealthy Release 2.1.0

jonnydawg:
Yes, thanks for noticing this! The test check and deployment check were updated preemptively for the 2.2.0 release.

jonnydawg:
The update to the test external check should no longer fatal!

PR #474, opened 2020-05-13
Test check example update

jonnydawg:
Removes the fatalln -- addresses #472

### DIFF
--- a/cmd/test-external-check/main.go
+++ b/cmd/test-external-check/main.go
@@ -46,19 +46,12 @@ func init() {
 	// Set check time limit to default
 	timeLimit = time.Duration(time.Minute * 10)
 	// Get the deadline time in unix from the env var
-	unixDeadline, err := checkclient.GetDeadline()
+	timeDeadline, err := checkclient.GetDeadline()
 	if err != nil {
 		log.Println("There was an issue getting the check deadline:", err.Error())
 	}
-	// Calculate check run duration from the deadline and now
-	deadline, err := strconv.Atoi(unixDeadline)
-	if err != nil {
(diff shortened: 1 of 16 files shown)


cmd/test-external-check/main.go
```diff
@@ -53,7 +53,7 @@ func init() {
 	// Calculate check run duration from the deadline and now
 	deadline, err := strconv.Atoi(unixDeadline)
```
integrii, line 54: I think this will break if ran and the environment variable does not exist still.  `deadline` here will equal 0, causing the `timeLimit` calculation to be 0 seconds.

jonnydawg: I set it to a default of 10 mins right before this

integrii: Oh, I see that now.  Looks almost good.

integrii:
Can you resolve conflicts (merge master into your branch) and then fix line `60`?

```
		timeLimit = time.Duration((int64(deadline) - time.Now().Unix()) * 1e9) // Multiply by 1,000,000,000 because that's how many nanoseconds are in a second
```

We should be able to do this without multiplying by 1 billion.

jonnydawg:
Yah, I can address this -- this doesn't need to happen anymore because the `GetDeadline` func returns a `time.Time{}` now instead of a string

jonnydawg:
Looks like that change got merged in with my deployment-check change from earlier? https://github.com/Comcast/kuberhealthy/blob/387bc304602c20364e755b79101b567b17611b72/cmd/test-external-check/main.go#L54

PR #482, opened 2020-05-19
Configmap

jdowni000:
Implemented a configmap in the place of having to set variables with flags. The new configmap is loaded from disk and is set with defaults. Additionally, I have updated the `README.md` to reflect the changes as well as the example to show the new `deployment.yaml` for example, and environment variable explanations that can be set.


deploy/helm/kuberhealthy/templates/deployment.yaml
```diff
@@ -31,6 +31,16 @@ spec:
...
       - image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+        volumeMounts:
+        - name: config-volume
+          mountPath: /etc/config/
+      volumes:
+       - name: config-volume
+         configMap:
+        # Provide the name of the ConfigMap containing the files you want
+        # to add to the container
+           name: special-config
```
integrii, line 42: the name of the configmap should not be `special-config`.  Let's make it simply `kuberhealthy`.  The type will be obvious.

deploy/helm/kuberhealthy/templates/configmap.yaml
```diff
@@ -30,4 +30,18 @@ data:
           description: Kuberhealthy shows that the cluster is not healthy
 {{- end -}}
 {{- end -}}
-
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: special-config
```
integrii, line 37: the name of the configmap should not be `special-config`.  Let's make it simply `kuberhealthy`.  The type will be obvious.

deploy/helm/kuberhealthy/templates/configmap.yaml
```diff
@@ -30,4 +30,18 @@ data:
...
+data:
+  kuberhealthy.yaml: |-
+    listenAddress: ":8080"
+    enableForceMaster: true
+    logLevel: "info"
+    influxUsername: ""
+    influxPassword: ""
+    influxURL: ""
+    influxDB: "http://localhost:8086"
+    enableInflux: false
```
integrii, line 47: Add some comments after each line here describing what it does.  You can use `#` to add a comment in this yaml

cmd/kuberhealthy/config.go
```diff
@@ -0,0 +1,70 @@
...
+
+	b, err := ioutil.ReadFile(file)
+	if err != nil {
+		return err
+	}
+
+	return yaml.Unmarshal(b, c)
+}
+
+// changeConfig creates a watcher and can be used to notify of change to configmap
```
integrii, line 36:
```suggestion
// configChangeNotifier creates a watcher and can be used to notify of change to the configmap file on disk
```

cmd/kuberhealthy/config.go
```diff
@@ -0,0 +1,70 @@
...
+func changeConfig() {
```
integrii, line 37:
```suggestion
func configChangeNotifier(configPath string) (chan struct{}) {
```

We want this `func` to return a channel that can be listed to in order to know when the file has changed.  The upstream code will monitor this channel and when a signal comes through, initiate reloading the main `Kuberhealthy` struct.

cmd/kuberhealthy/config.go
```diff
@@ -0,0 +1,70 @@
...
+				if !ok {
+					return
+				}
+				log.Infoln("event:", event)
+				if event.Op&fsnotify.Write == fsnotify.Write {
+					log.Infoln("modified file:", event.Name)
+				}
+			case err, ok := <-watcher.Errors:
+				if !ok {
+					return
```
integrii, line 58:
```suggestion
					log.Infoln("Failed to watch", configPath, "with error", err)
```

returning here would exit the background go routine watching the file

cmd/kuberhealthy/config.go
```diff
@@ -0,0 +1,70 @@
...
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				log.Infoln("event:", event)
+				if event.Op&fsnotify.Write == fsnotify.Write {
+					log.Infoln("modified file:", event.Name)
```
integrii, line 54: instead of logging here, you would notify the channel that gets returned to the user:

```
notifyChan <- struct{}{}
```

cmd/kuberhealthy/config.go
```diff
@@ -0,0 +1,70 @@
...
+				log.Infoln("error:", err)
+			}
+		}
+	}()
+
+	err = watcher.Add(configPath)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	<-done
```
integrii, line 69: instead of watching for done, you should return the channel that notifies the user so they can listen to it

cmd/kuberhealthy/config.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// changeConfig creates a watcher and can be used to notify of change to configmap
+func changeConfig() {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Fatalln(err)
+	}
+	defer watcher.Close()
+
```
integrii, line 43: I would make a new channel here to return to the user...

```
notifyChan := make(chan struct{})
```

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -146,7 +146,7 @@ func (k *Kuberhealthy) Start(ctx context.Context) {
 	go k.stateReflector.Start()
 
 	// if influxdb is enabled, configure it
-	if enableInflux {
+	if cfg.EnableInflux == false {
```
jonnydawg, line 149: This would turn on influx if EnableInflux is false no?

jdowni000: That is correct...I have pushed an update that checks for true....I am not 100% positive why I did that at the time

deploy/helm/kuberhealthy/templates/deployment.yaml
jdowni000: Changes made with 6bf8b54

deploy/helm/kuberhealthy/templates/configmap.yaml
jdowni000: Changes made with 6bf8b54

deploy/helm/kuberhealthy/templates/configmap.yaml
jdowni000: Changes made with 6bf8b54

cmd/kuberhealthy/config.go
jdowni000: Changes made with 6bf8b54

cmd/kuberhealthy/config.go
jdowni000: Changes made with 6bf8b54

cmd/kuberhealthy/config.go
jdowni000: Changes made with 6bf8b54

cmd/kuberhealthy/main.go
```diff
@@ -45,12 +44,14 @@ var listenNamespace string         // namespace to listen (watch/get) `khcheck`
 var terminationGracePeriod = time.Minute * 5 // keep calibrated with kubernetes terminationGracePeriodSeconds
 
 // flags indicating that checks of specific types should be used
-var enableForceMaster bool // force master mode - for debugging
-var enableDebug bool       // enable debug logging
-var logLevel = "info"
+var DSPauseContainerImageOverride string // specify an alternate location for the DSC pause container - see #114
+// DSTolerationOverride specifies an alternate list of taints to tolerate - see #178
+var DSTolerationOverride []string
```
joshulyne, line 49: hmm can you check is this is being used in the kuberhealthy project anywhere? pretty sure this was for the daemonset check and thats now in a different package!

cmd/kuberhealthy/main.go
```diff
@@ -45,12 +44,14 @@ var listenNamespace string         // namespace to listen (watch/get) `khcheck`
...
 
 // the hostname of this pod
 var podHostname string
+var enablePodStatusChecks = determineCheckStateFromEnvVar("POD_STATUS_CHECK")
+var enableExternalChecks = true
```
joshulyne, line 54: ^^hmm same here, not sure if these vars `enablePodStatusChecks`, `enableExternalChecks` are needed anymore

PR #512, opened 2020-06-04
Check-reaper fix to keep failed pods around

joshulyne:


cmd/check-reaper/main.go
```diff
@@ -135,7 +135,7 @@ func deleteFilteredCheckerPods(client *kubernetes.Clientset, reapCheckerPods map
 			}
 
 			// Delete if there are more than 5 checker pods with the same name in status Succeeded that were created more recently
-			if successOldCount > MaxPodsThreshold && successCount > MaxPodsThreshold {
+			if  v.Status.Phase == "Succeeded" && successOldCount > MaxPodsThreshold && successCount > MaxPodsThreshold {
```
jonnydawg, line 138: Instead of hard-coding `"Succeeded"` I think you should use the const from the k8s [corev1 pkg](https://godoc.org/k8s.io/api/core/v1#PodPhase)

jonnydawg: Same with line `150` below

Issue #473, opened 2020-05-13
Pod eviction only looking for Running pods when cleaning up KH checks, leading to potential memory usage issues.


zjhans:
**Describe the bug**
Pod eviction currently appears to only target pods tagged with a status.phase=Running. This can lead to continuous spawning of pods with a status of ImagePullBackOff or another error state, resulting in a node crash due to resource/memory usage.

Logs:
time="2020-05-12T17:18:17Z" level=info msg="c2f17ef8-f86f-4e7f-9a42-ff045d96a672 kuberhealthy/kh-external-test-check: [eviction: looking for pods with the label kuberhealthy-check-name = kh-external-test-check and status.phase=Running]"

**Steps To Reproduce**
- Deploy a test check that will intentionally fail (I.e., point to a repository that does not exist, set ImagePullPolicy to Always
- Leave the khcheck active, failed pods will continue to accumulate.

**Expected behavior**
I believe all old, unneeded pods for the khcheck in question should be evicted.

**Screenshots**
[Uploading image.png…]()

**Versions**
 - Cluster OS: Docker Desktop
 - Kubernetes Version: 1.16.6
 - Kuberhealthy Release 2.1.0

jonnydawg:
Thanks for bringing this up! There have been reports of memory leaks, but we have not been able to reproduce it.

zjhans:
Original image failed to upload.

![image](https://user-images.githubusercontent.com/55612469/81934462-fbda4f00-95bc-11ea-9a19-62623c3f2fcb.png)

Issue #221, opened 2019-10-15
Create an image pull test


integrii:
Sometimes upstream repositories go down.  It would be nice to have an alert for this because it could quickly turn into a production problem.  All we need to do is configure a `khcheck` that uses the example checker and an `imagePullPolicy` of `Always`.

Docker went down today and Kuberhealthy v2 showed errors in our test lab, but it would be nice to have a dedicated check for this.

sladyn98:
Hello, I would like to take up this issue in order to get familiar with the codebase, could you help me begin?

integrii:
Hey there @sladyn98!  We would love to have your contribution.  This check should be easy.  Here is what I am thinking:

- Create a checker pod that ALWAYS exits by calling `checkclient.ReportSuccess()`
- Create a check spec (khcheck yaml) that uses this new check image and has `imagePullPolicy` set to `Always`

I would create the check in the `cmd/` directory of this project (fork the repo so you have access).  You can use [this example check](https://github.com/Comcast/kuberhealthy/tree/master/cmd/test-external-check) as a starting point (just copy the entire directory to get started).

When it comes time to make the check official, just push your image somewhere public (like a free quay.io account) and make a PR to put it in the [check registry](https://github.com/Comcast/kuberhealthy/blob/master/docs/EXTERNAL_CHECKS_REGISTRY.md).

Here are the docs for making your own check: https://github.com/Comcast/kuberhealthy/blob/master/docs/EXTERNAL_CHECK_CREATION.md

PR #516, opened 2020-06-09
Image pull checker and eviction fix

zjhans:
This commit resolves issues [#221](https://github.com/Comcast/kuberhealthy/issues/221) and [#473](https://github.com/Comcast/kuberhealthy/issues/473) with the following updates:

- Adds an external image pull check to test-external-check
- Adds more in depth PodKill and PodNameExists functions to util package
- Updates cleanup function:
    -Adds error return to evictPod function
    -Checks for pods that are not in Failed or Succeeded state, instead of only Running state.
    -If evictPod returns an error, run util.PodNameExists and if it returns true, run util.PodKill, 
      returning any errors

### DIFF
--- a/cmd/kuberhealthy/util.go
+++ b/cmd/kuberhealthy/util.go
@@ -40,7 +40,7 @@ func getAllLogLevel() string {
 }
 
 // notifyChanLimiter takes in a chan used for notifications and smooths it out to at most
-// one single notification every specifid duration.  This will continuously empty whatever the inChan
+// one single notification every specified duration.  This will continuously empty whatever the inChan
 // channel fed to it is.  Useful for controlling noisy upstream channel spam. This smooths notifications
 // out so that outChan is only notified after inChan has been quiet for a full duration of
 // the specified maxSpeed.  Stops running when inChan closes
@@ -49,17 +49,24 @@ func notifyChanLimiter(maxSpeed time.Duration, inChan chan struct{}, outChan cha
 	// we wait for an initial inChan message and then watch for spam to stop.
 	// when inChan closes, the func exits
 	for range inChan {
-		log.Println("channel notify limiter witnessed an upstream message on inChan")
+		log.Infoln("channel notify limiter witnessed an upstream message on inChan")
+
+		// Label for following for-select loop
+	notifyChannel:
 		for {
+			log.Debugln("channel notify limiter waiting to receive another inChan or notify after", maxSpeed)
 			select {
 			case <-time.After(maxSpeed):
+				log.Debugln("channel notify limiter reached", maxSpeed, ". Sending output")
 				outChan <- struct{}{}
-				break
+				// break out of the for-select loop and go through next inChan loop iteration if any.
+				break notifyChannel
 			case <-inChan:
-				log.Println("channel notify limiter witnessed an upstream message on inChan and is waiting an additional", maxSpeed, "before sending output")
+				log.Debugln("channel notify limiter witnessed an upstream message on inChan and is waiting an additional", maxSpeed, "before sending output")
 			}
-			return
 		}
+
+		log.Debugln("channel notify limiter finished going through notifications")
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/checks/external/util/main.go
```diff
@@ -88,4 +89,48 @@ func GetInstanceNamespace(defaultNamespace string) string {
 	}
 
 	return instanceNamespace
-}
\ No newline at end of file
+}
+
+// PodNameExists determines if a pod with the specified name exists in the specified namespace.
+func PodNameExists(client *kubernetes.Clientset, podName string, namespace string) (bool, error) {
```
joshulyne, line 95: so we have another function `podExists` in this external/main.go file that almost does the same thing:
https://github.com/Comcast/kuberhealthy/blob/master/pkg/checks/external/main.go#L1189-L1216. Could we find where this function `podExists` is being used and replace with this one?

cmd/kuberhealthy/util.go
```diff
@@ -54,9 +54,11 @@ func notifyChanLimiter(maxSpeed time.Duration, inChan chan struct{}, outChan cha
 			select {
 			case <-time.After(maxSpeed):
 				outChan <- struct{}{}
+				break
 			case <-inChan:
 				log.Println("channel notify limiter witnessed an upstream message on inChan and is waiting an additional", maxSpeed, "before sending output")
 			}
+			return
```
joshulyne, line 61: this was old / temp fix! i believe you can now pull / merge with what's in master!

cmd/test-external-check/README.md
```diff
@@ -45,3 +45,46 @@ spec:
           cpu: 10m
           memory: 50Mi
 ```
+
+### image-pull-check
+
+This is a container used to test availability of external image respositories in Kuberhealthy.  Simply place the `kuberhealthy/test-external-check` image on the repository you need tested.
```
joshulyne, line 51: `This is a container used to test availability of external image respositories in Kuberhealthy.` --> `This container tests the availability of external image repositories using Kuberhealthy`.

cmd/test-external-check/README.md
```diff
@@ -45,3 +45,46 @@ spec:
...
+
+##### Example `khcheck`
+
+The pod is designed to attempt a pull of the test image from the remote repository (never from local) every 10 minutes. If the image is unavailable, an error will be reported to the Kuberheakthy API.
+
+To put a copy of this image to your repository, run `docker pull kuberhealthy/test-external-check` and then `docker push my.repository/repo/test-external-check`.
```
joshulyne, line 57: can this comment be placed in the `### image-pull-check` section? feel like it might better fit there after `Simply place the `kuberhealthy/test-external-check` image on the repository you need tested.`

integrii:
@zjhans check the conflicts here and re-submit the PR... I really like your detailed and concise github descriptions.

pkg/checks/external/main.go
```diff
@@ -253,38 +253,57 @@ func (ext *Checker) getCheck() (*khcheckcrd.KuberhealthyCheck, error) {
...
+						ext.log("forcefull killing pod", p.GetName, "in 15 seconds")
+						err := util.PodKill(ext.KubeClient, p.GetName(), p.GetNamespace(), 15)
+						if err != nil {
+							ext.log("error killing pod", p.GetName()+":", err)
+						}
+					}
+				}
+				wg.Done()
+			}(p)
+		}
```
integrii, line 300: Can we flatten this out or break it into a few supporting functions?

zjhans: I removed redundant logging since the evictPod, PodKill, and PodNameExists functions already do error logging and handling. The only other option I could think of was to combine PodNameExists and PodKill into the evictPod function but that seems it would defeat the purpose of breaking the functions out in the first place.

zjhans: Calls to util.PodNameExists and util.PodKill are now part of the evictPod func as discussed.

Issue #518, opened 2020-06-11
Gracefully handle new node startup race conditions


integrii:
It is possible for a checker pod to be scheduled and started on a node before the node is fully ready.  This makes the check instantly fail.  Most commonly, this happens when the checker pod starts before `kube-proxy` has had a chance to apply netfilter rules.

I found [this relevant issue](https://github.com/kubernetes/kubernetes/issues/75890) that describes the issue and shows how some people solved the problem by creating their own operator to un-taint nodes when they are fully ready.

This seems like an oversight with Kubernetes fundamentals and is disappointing.  This thread should be about how we want to work around this, or pursue fixes with the Kubernetes project directly.

joshulyne:
For our faster / more frequently run checks, ie dns-resolution-check -- we added a temp fix: #519 to make the check sleep for an extra minute IF the node age is less than 3m. A better fix would be to not just check the node age but also check to see if kube-proxy on that node is ready and running before running the check. This functionality can be a package that other faster / more frequently run checks can implement.

PR #519, opened 2020-06-11
Check node age when running dns-resolution check

joshulyne:
Partly addresses https://github.com/Comcast/kuberhealthy/issues/518. DNS resolution check runs quickly and often (every 2m by default). When this check comes up on a new node, check the node age and if it's less than 3 minutes old, wait 1 minute before running the check. We see new node failures mostly for this check since it runs often/quickly. Wondering if we can even extend this logic for all checks if needed and as a temporary fix until this issue is resolved: https://github.com/kubernetes/kubernetes/issues/75890

### DIFF
--- a/cmd/dns-resolution-check/Makefile
+++ b/cmd/dns-resolution-check/Makefile
@@ -1,5 +1,5 @@
 build:
-	docker build -t kuberhealthy/dns-resolution-check:v1.2.0 -f Dockerfile ../../
+	docker build -t kuberhealthy/dns-resolution-check:v1.4.0 -f Dockerfile ../../
 
 push:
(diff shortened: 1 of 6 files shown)


integrii:
This seems like a good first mitigation to the issue.  A way to improve this would be to check if `kube-proxy` has been running for some period of time on the node. 

I'm approving for now!

joshulyne:
> This seems like a good first mitigation to the issue. A way to improve this would be to check if `kube-proxy` has been running for some period of time on the node.
> 
> I'm approving for now!

we should add that to your issue  #518 --> in the `checkNodeAge` there should also be a `checkKubeProxy` to see if kube-proxy is ready / has been running on the node for a bit

PR #523, opened 2020-06-12
Add python-client

bbkgh:
Python-client has been added by following JavaScript client codes.
It's better to make a PyPi library from it and publish that by your (kuberhealthy) credentials for making it more easier to use.

Signed-off-by: Babak Ghadiri <bghadiri@hamravesh.com>


integrii:
This is great.  Thank you for contributing to Kuberhealthy and adding a nice example for Python programmers to get started.  

You may want to also add a check [on the writing your own checks page](https://github.com/Comcast/kuberhealthy/blob/master/docs/EXTERNAL_CHECK_CREATION.md).

@jonnydawg is taking a look more deeply at this for feedback.

clients/python/kh_client.py
```diff
@@ -0,0 +1,61 @@
+import dataclasses
+import json
+import os
+from dataclasses import dataclass
+from typing import List
+
+import requests
```
jonnydawg, line 7: Could this package be included in a `requirements.txt` file -- I think this would help with installing dependencies for this package such as requests.

clients/python/example_check.py
```diff
@@ -0,0 +1,20 @@
+from .kh_client import *
+import os
+
+fail = os.getenv("fail", False)
```
jonnydawg, line 4: Environment variables are typically uppercase, could you uppercase this env var?

clients/python/example_check.py
```diff
@@ -0,0 +1,20 @@
+from .kh_client import *
```
jonnydawg, line 1: Since the `kh_client` is in the same directory you don't need the `.` -- this causes python to misinterpret the location of the import, creating a `ModuleNotFoundError`.

clients/python/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:3.7
+COPY . .
```
jonnydawg, line 2: I think we need to copy a `requirements.txt` in this directory to help with missing python dependencies in the container like `requests`. We would just need to also copy the `requirements.txt` into the container and run an install `pip3 install -r requirements.txt`.

clients/python/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+```shell
+cd my-kh-check
+curl -O -L https://raw.githubusercontent.com/Comcast/kuberhealthy/master/clients/python/kh_client.py
+```
+
+In your project, require the file you just downloaded:
+
+```python
+from .kh_client import *
```
jonnydawg, line 19: In regards to the other comment, would this line need to be updated?

PR #537, opened 2020-06-20
fix goroutine leak in watchForCheckerPodShutdown()

### LINKED ISSUE #533: Possible goroutines leak from watch events

sbueringer:
Fixes #533

I fixed the leak on a heavily modified branch wit logging of goroutine ids including who creates which goroutine etc. This is the change which finally fixed the leak. I'm currently verifying that this version without all other changes also fixes the leak.

### DIFF
--- a/pkg/checks/external/main.go
+++ b/pkg/checks/external/main.go
@@ -402,6 +402,11 @@ func (ext *Checker) watchForCheckerPodShutdown(shutdownEventNotifyC chan struct{
 	defer func() {
 		if watcher != nil {
 			watcher.Stop()
+			// Empty the resultChan after stopping.
+			// Otherwise it can happen that the  StreamWatcher is blocked when it tries
+			// to put an Event in the result channel. This happens when no consumer exists
+			// anymore to empty the ResultChan after calling Stop().
+			for range watcher.ResultChan() {}
 		}
 	}()
 


sbueringer:
@integrii Could it be that you dropped the capability to set the kubecfg on master? We used that with the 2.2.0 version. I saw that the flag was removed and that there's a property in the config file. But I'm not sure how to use that as it's a unexported variable and therefore not parsed? Looks to me like I have to overwrite the HOME var in my Pod and put the kubeconfig under $HOME/.kube/config?

integrii:
Ah yeah. The new `$KUBECONFIG` env var should work.  I can test sometime soon... This is part of the new configmap and live reloading that was just merged to master today.  It's still due for some more testing before it gets into a release...

integrii:
By the way, I plan to review and test this PR Monday... I really appreciate your debugging so carefully to find root cause.  Awesome!

sbueringer:
> Ah yeah. The new `$KUBECONFIG` env var should work. I can test sometime soon... This is part of the new configmap and live reloading that was just merged to master today. It's still due for some more testing before it gets into a release...

Ah thx. Mounting it at ~/.kube/config also worked :)

sbueringer:
> By the way, I plan to review and test this PR Monday... I really appreciate your debugging so carefully to find root cause. Awesome!

Thx, no problem :). Looks like it works in my env still 6Mi memory usage after 6 hours. Usually it was already higher at this point. Btw I assume the goroutine leak is a race condition which is not always triggered, e.g. when I run kuberhealthy outside the cluster in my IDE I cannot reproduce it.

cagdascirit:
@sbueringer @integrii I have patched v2.2.0 with this commit and the memory leak problem has been resolved in our test environment. For 10+ hours memory consumption kept steady

integrii:
Thanks for testing guys. I have a branch I'm going to PR tomorrow that has a slightly different take on this. Instead of emptying the channel, we leave the for loop running until the range breaks due to a closed channel. This ensures that the channel is always emptied. 

I also chased down a couple other missing channel closes that were needed and simplified the retry logic...

sbueringer:
> Thanks for testing guys. I have a branch I'm going to PR tomorrow that has a slightly different take on this. Instead of emptying the channel, we leave the for loop running until the range breaks due to a closed channel. This ensures that the channel is always emptied.
> 
> I also chased down a couple other missing channel closes that were needed and simplified the retry logic...

Yup sure, sgtm. I'm aware my "solution" is just a quick hack but doesn't solve the underlying problem (that the consumer of the streamwatcher is terminated before stop is called)

integrii:
I have refactored this in the branch `pr-fix-goroutine-leak` (bc22bd1) here and am testing locally before opening a PR...  This fix includes a refactor to the channel handling.  To solve the problem identified in this thread, the reader of the watch channel does not stop reading until the channel is closed, but to prevent an upstream blockage, only writes one message back to the original caller.

This branch also includes a bunch of other readability and simplification...

integrii:
Feel free to test this build while I am working on getting a suitable test environment handy: `docker pull kuberhealthy/kuberhealthy:537`

integrii:
Remember to apply this yaml configmap now that configmap support is in master... Be sure to use the updated `deployment` and `configmap` from [this spec](https://github.com/Comcast/kuberhealthy/blob/master/deploy/kuberhealthy.yaml)

integrii:
I've got metrics on this now... I'll let you guys know how it looks in a day.

image

integrii:
Memory is looking good!  This is the master pod:
image

Here is the secondary pod. looks like it's hardly using any memory:
image

cagdascirit:
@integrii which branch we can use to collect the fixes?

sbueringer:
@integrii When there's another PR ready which fixes this, I'm happy to close this PR and test your fix in our env. For now, I'm not really in a hurry as we're using my patch from this PR in production and it looks fine.

integrii:
Checked today and this memory leak is still there on my branch it looks like.  I will revisit soon.

image

Issue #447, opened 2020-04-29
Another serious bug, the main process is blocked


calmkart:
**Describe the bug**
This bug is similar to #439 , but not the same, it is very tricky

How to reproduce this bug
1. create a khcheck cr
2. the khcheck  cr will create a kh check pod 
3. if the check pod keep containerCreating (any reason)
4. kuberhealhty pod main process blocked

Please note that although this bug looks similar to #439 , it is completely different. #439  bugs are complicated to troubleshoot, but simple to solve. This bug is very easy to reproduce, but it is more difficult to solve. Below I will explain in detail the cause of the bug.

If we make any changes to khcheck cr, the event will be monitored, and a channel is issued without distinguishing the event type
https://github.com/Comcast/kuberhealthy/blob/7572e86a2b433864bc7d647203496a8411ca5dc5/cmd/kuberhealthy/kuberhealthy.go#L295-L314
Then restart all khcheck
https://github.com/Comcast/kuberhealthy/blob/7572e86a2b433864bc7d647203496a8411ca5dc5/cmd/kuberhealthy/kuberhealthy.go#L195-L199
but in `k.StopChecks()` we should wait for all check pod shutdown
https://github.com/Comcast/kuberhealthy/blob/7572e86a2b433864bc7d647203496a8411ca5dc5/cmd/kuberhealthy/kuberhealthy.go#L122-L132

This time a bug appeared
Our pod has not started at all, it has been stuck in the containerCreating state, waitforstart () can not be completed, it is impossible to complete shutdown ()

How to fix this bug is worth thinking about. I think it is impossible to fix this kind of bug with the logic of restarting all khcheck by monitoring any changes at present. Because we can't judge whether the pod is in ContainerCreating, it is really dead or busy (for example, in pull image), so we can't kill these pods automatically when restarting, we can only wait for the pods to automatically recycle after the timeout period. In this process, the main process is stuck.

I think a good way to solve this kind of problem (or even other bugs) is to distinguish the type of khcheck change event that is monitored, and do different treatments for different types of events, instead of restarting the simple and rough monitoring events like now, including The processing of khstate is also rough.

[...]

PR #540, opened 2020-06-25
Address bug 447. Modify eviction for pods not in failed / succeeded phase.

joshulyne:
Addresses #447.

When a khcheck pod is stuck in pending or "container creating" this blocks the main kuberhealthy process. The eviction policy modification added in pr #516 addresses the state when khcheck pods are stuck in imagepullerror status, but does not cover "pending" status. Added a select and timeout during the waitForShutdown process in the external main.go package so that the main kuberhealthy process doesn't get stuck until the check timeout or next run iteration. 

Also, modified eviction filtering for running, pending, or unknown pods instead of excluding succeeded or failed pods. Seems like ` p.Status.Phase != apiv1.PodFailed || p.Status.Phase != apiv1.PodSucceeded` wasn't working as expected and all of that khchecker pods were getting evicted instead of evicting just the nonfailed/succeeded pods.

### DIFF
--- a/pkg/checks/external/main.go
+++ b/pkg/checks/external/main.go
@@ -1209,15 +1209,17 @@ func (ext *Checker) setNewCheckUUID() error {
 }
 
 // waitForShutdown waits for the external pod to shut down
-func (ext *Checker) waitForShutdown(ctx context.Context) error {
+func (ext *Checker) waitForShutdown(ctx context.Context) chan error {
 	// repeatedly fetch the pod until its gone or the context
 	// is canceled
+	doneChan := make(chan error, 1)
 	for {
 		time.Sleep(time.Second * 5)
 		exists, err := util.PodNameExists(ext.KubeClient, ext.checkPodName, ext.Namespace)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/checks/external/main.go
```diff
@@ -1238,21 +1242,33 @@ func (ext *Checker) Shutdown() error {
...
-		ext.log("Error waiting for pod removal during shutdown:", err)
-		return err
-	}
-
-	// wait for all background checkers and workers to finish before the check is fully "shutdown"
-	ext.log("Waiting for background workers to cleanup...")
-	ext.wg.Wait()
+	go func() {
+		log.Debugln("Waiting for pod", ext.podName(), "to shutdown")
+		doneChan <- ext.waitForShutdown(ctx)
```
integrii, line 1252: Why not just put this directly in the select below on line 1256?  It seems like all this go routine adds is a couple logs.  Can we remove `doneChan` completely here?

joshulyne: yupp but i'll have to modify `waitForShutdown` to return a `chan error`

PR #543, opened 2020-06-26
Set timeout for context cancel with timeout greater than check time l…

joshulyne:
…imit

Found a bug where ctxCancel has the same timeout as checkTimeLimit so the ctx.Done() is always called at the same time as  time.After(CheckTimeLimit) during our `waitFor...`

### DIFF
--- a/cmd/daemonset-check/input.go
+++ b/cmd/daemonset-check/input.go
@@ -17,6 +17,9 @@ func parseInputValues() {
 	if err != nil {
 		log.Infoln("There was an issue getting the check deadline:", err.Error())
 	}
+	checkTimeout = timeDeadline.Sub(time.Now().Add(time.Second * 5))
+	log.Infoln("Check time out set to:", checkTimeout)
+
 	// Give 60 seconds to clean up / shut down once check time limit is reached before the kh deadline
 	checkTimeLimit = timeDeadline.Sub(time.Now().Add(time.Second * 60))
 	log.Infoln("Setting check time limit to:", checkTimeLimit)
--- a/cmd/daemonset-check/input.go
+++ b/cmd/daemonset-check/input.go
@@ -11,18 +11,32 @@ import (
 // parseInputValues parses and sets global vars from env variables and other inputs
 func parseInputValues() {
 
-	// Use injected pod variable KH_CHECK_RUN_DEADLINE to set check time limit
-	checkTimeLimit = defaultCheckTimeLimit
-	timeDeadline, err := kh.GetDeadline()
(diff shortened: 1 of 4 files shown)


cmd/daemonset-check/input.go
```diff
@@ -17,6 +17,9 @@ func parseInputValues() {
 	if err != nil {
 		log.Infoln("There was an issue getting the check deadline:", err.Error())
 	}
+	checkTimeout = timeDeadline.Sub(time.Now().Add(time.Second * 5))
```
integrii, line 20: I think we should better identify what our time barriers are here...

- `checkTimeLimit` is how long we want this pod to run before considering the check a failure
- `timeDeadline` is the point when the checker pod is "dead" in the eyes of KH. no status can be reported after this time.
- `checkTimeout` is when the check officially considers the check as a failure and begins reporting it

This seems a little too complex.  We also have a `defaultCheckTimeLimit` variable and a `defaultShutdownGracePeriod`.  



How about we simplify the times here into this:

- the time at which the check will be orphaned by Kuberhealthy (`timeDeadline`)
- the time at which the check needs needs to consider its self failed and report a failure 

The orphaned time is simply `kh.GetDeadline()`.  The time at which the check needs to fail is that time, minus the graceful shutdown period.  Something like:

```
shutdownGracePeriod := time.Minute
podDeadline, err := kh.GetDeadline()
if err != nil {
    // stuff
}
checkDeadline := podDeadline.Add(-shutdownGracePeriod)
```

perhaps we could simplify to this?

integrii: After you have `checkDeadline`, you can do stuff like this:

```go
if time.Now().After(checkDeadline) {
    // failure
}
```

joshulyne: sounds good!

cmd/daemonset-check/run_check.go
```diff
@@ -114,8 +114,8 @@ func remove(dsName string, ctx context.Context) error {
 			return fmt.Errorf("error trying to delete daemonset: %s", err)
 		}
 		log.Infoln("Successfully requested daemonset removal.")
-	case <-time.After(checkTimeLimit):
-		return errors.New("Reached check pod timeout: " + checkTimeLimit.String() + " waiting for daemonset removal command to complete.")
+	case <-time.After(checkDeadline.Sub(now)):
```
integrii, line 117: for each of these, you could receive the channel from a single `time.After` call and then just read from it in the case:

```go
deadlineChan := time.After(checkDeadline.Sub(now))
```

then

```go
  case <-deadlineChan:
```

PR #544, opened 2020-06-29
Add nodeCheck package

joshulyne:
Addresses #518 

Add a nodeCheck package that provides two checks that kuberhealthy checks can so that fast running checks don't quickly fail on new nodes. 

Both checks initially check the age of the node which the khcheck pod is running on. If the node is less than 3 minutes old, check either if kube proxy is ready and running on the node OR if the kuberhealthy endpoint can be hit. 
`WaitForKuberhealthyEndpoint` --> checks if kuberhealthy endpoint can be hit
`WaitForKubeProxy` --> checks if kube proxy is ready and running on the node


go.mod
```diff
@@ -33,13 +27,40 @@ require (
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/ini.v1 v1.51.0 // indirect
 	gopkg.in/yaml.v2 v2.2.8
-	k8s.io/api v0.0.0-20190819141258-3544db3b9e44
-	k8s.io/apimachinery v0.0.0-20190817020851-f2f3a405f61d
-	k8s.io/client-go v0.0.0-20190819141724-e14f31a72a77
+	k8s.io/api v0.0.0
+	k8s.io/apimachinery v0.0.0
+	k8s.io/client-go v0.0.0
```
integrii, line 32: We're sneaking in a big change to our module dependencies here... can we do this without tampering with the go module so much?   Can we leave these pinned to their prior revisions and make things build still?

pkg/checks/external/nodeCheck/main.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+			log.Infoln("Kube proxy is ready. Proceeding to run check.")
+		case <- time.After(time.Duration(time.Minute)):
+			// Just log the error.
+			// TO DO: figure out how to address this. Should the check to skip this run and pass up an error instead?
+			// If kube-proxy isn't ready and running, there's definitely something wrong with the new node coming up.
+			log.Errorln("Timed out checking if kube proxy is ready. Check node:", nodeName,
+			"Check may or may not complete successfully.")
+		}
+		return
```
integrii, line 50:
```suggestion
```

pkg/checks/external/nodeCheck/main.go
```diff
@@ -0,0 +1,99 @@
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/pkg/api/v1/pod"
+)
+
+// checkNode checks the node's age to make sure it's not less than three minutes old. If so, sleep for one minute
+// and check if kube proxy is ready and running on the node before running the check. This function requires the
+// khcheck to expose the pod's node name information using environment variables.
+func CheckNode(client *kubernetes.Clientset, nodeName string, now time.Time) {
```
integrii, line 17: How about the caller passes in the pod name (and maybe namespace) instead of the node name?  What if we just go one step further and assume the pod's hostname is the result of `os.Hostname()`?

Also, can we rename `CheckNode` here to something more descript?  The current call is `nodeCheck.CheckNode()` which is very generic.  Perhaps rename to `nodeCheck.WaitForKubeProxy`?

pkg/checks/external/nodeCheck/main.go
```diff
@@ -0,0 +1,99 @@
...
+		time.Sleep(time.Second * 5)
+	}
+}
+
+// checkKubeProxyPod gets the kube proxy pod and checks if its ready and running.
+func checkKubeProxyPod(client *kubernetes.Clientset, podName string) (bool, error) {
+
+	var kubeProxyReady bool
+
+	kubeProxyPod, err := client.CoreV1().Pods("kube-system").Get(podName, v1.GetOptions{})
```
integrii, line 83: We should not assume `kube-system` is used here... Perhaps we could just query for `kube-proxy` pods  that are on this node using a selector.

pkg/checks/external/nodeCheck/main.go
integrii, line 17: How about the caller passes in the pod name (and maybe namespace) instead of the node name?  What if we just go one step further and assume the pod's hostname is the result of `os.Hostname()`?

Also, can we rename `CheckNode` here to something more descript?  The current call is `nodeCheck.CheckNode()` which is very generic.  Perhaps rename to `nodeCheck.WaitForKubeProxy`?  I know that we have a `waitForKubeProxyReady` func, but that isn't exposed to the user for calling.

go.mod
integrii, line 32: We're sneaking in a big change to our module dependencies here... can we do this without tampering with the go module so much? Can we leave these pinned to their prior revisions and make things build still?

pkg/checks/external/nodeCheck/main.go
```diff
@@ -0,0 +1,203 @@
...
+		// Just log the error and return since this check will not work if there's an error getting the node
+		log.Errorln("Error checking if node is new:", khPod.Spec.NodeName, err)
+		return
+	}
+
+	if nodeNew {
+		sleepDuration := time.Minute
+		log.Infoln("Node is less than", nodeMinAge, "old. Sleeping for", sleepDuration)
+		time.Sleep(sleepDuration)
+
```
joshulyne, line 75: not sure if we still want to sleep for a minute before checking kube-proxy. We time out after a minute waiting for kube-proxy to be ready and running.

pkg/checks/external/nodeCheck/main.go
```diff
@@ -0,0 +1,203 @@
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/pkg/api/v1/pod"
+)
+
+// WaitForKuberhealthyEndpoint first checks the node's age to make sure it's not less than three minutes old. If so, it
+// checks if the kuberhealthy endpoint (KH_REPORTING_URL) can be hit, determining whether or not the proper netfilter
+// rules have been set up on the node in order to hit the kuberhealthy service.
+func WaitForKuberhealthyEndpoint(client *kubernetes.Clientset, namespace string, kuberhealthyEndpoint string) {
```
integrii, line 19: Perhaps we don't wait 3 minutes here... simply scope this down to just checking the KH endpoint?  If they want to wait 3 minutes, or wait for the kube-proxy pod to come up, they can use the other funcs added here.

pkg/checks/external/nodeCheck/main.go
```diff
@@ -0,0 +1,203 @@
...
+
+	khPod, err := getKHPod(client, namespace)
+	if err != nil {
+		// Just log the error and return since this check will not work if there's an error getting the podName
+		log.Errorln("Error getting kuberhealthy pod:", err)
+		return
+	}
+	log.Infoln("Pod is on node:", khPod.Spec.NodeName)
+
+	nodeMinAge := time.Minute * 3
```
integrii, line 63: same thing here - perhaps let the user check the node age on their own?

pkg/checks/external/nodeCheck/main.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		// Just log the error and return since this check will not work if there's an error getting the pod
+		log.Errorln("Error getting pod:", err)
+		return khPod, err
+	}
+	return khPod, err
+}
+
+// checkIfNodeIsNew checks the age of the node the kuberhealthy check pod is on to determine if its "new" or not.
+func checkNodeNew(client *kubernetes.Clientset, khPod *corev1.Pod, nodeMinAge time.Duration) (bool, error) {
```
integrii, line 117: we could export this one so users can use ut, probably... Maybe make the exported function take in a pod name and namespace instead of a `corev1.Pod`?

integrii:
The three main calls in the package are looking good.  Here are some thoughts:

- We shouldn't log anything from these packages because we shouldn't assume that its okay to log in a specific format or to a specific output.  We should return errors only.
- We might want to remove the static one minute timeouts in here.  If the user supplies a context, they are able to cancel at any time, so we don't need to additional timeout.
- We should pass the context supplied by the user down to anything that has a for loop (like `waitForKuberhealthyEndpointReady`) so that those routines don't keep running after the user revokes their context.

pkg/checks/external/nodeCheck/main.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		log.Debugln("Kube proxy is ready. Proceeding to run check.")
+	case <-ctx.Done():
+		return errors.New("context cancelled waiting for kube proxy to be ready and running")
+	}
+	return nil
+}
+
+// WaitForNodeAge checks the node's age to see if its less than the minimum node age. If so, sleeps for a given sleep duration.
+func WaitForNodeAge(client *kubernetes.Clientset, namespace string, minNodeAge time.Duration, sleepDuration time.Duration, ctx context.Context) error {
```
integrii, line 70: Lets make this check repeatedly ask if the node is new until it's not.  That way, when you call it, it will block until the node is old enough.  Don't worry about sleeping here.  The user can do that in the calling code.

joshulyne: instead of asking the node repeatedly if its new or not -- can we just sleep until it reaches the minimum node age?

integrii: Ah yeah, then re-check.  Good idea.

- fetch node age
- calculate when node should be old enough
- sleep the remaining duration
- re-check node age after sleeping? (maybe not)

joshulyne: imo idt we need to re-check? unless we want to be SURE its past its minNodeAge...?

pkg/checks/external/nodeCheck/main.go
```diff
@@ -66,29 +66,34 @@ func WaitForKubeProxy(client *kubernetes.Clientset, namespace string, ctx contex
 	return nil
 }
 
-// WaitForNodeAge checks the node's age to see if its less than the minimum node age. If so, sleeps for a given sleep duration.
-func WaitForNodeAge(client *kubernetes.Clientset, namespace string, minNodeAge time.Duration, sleepDuration time.Duration, ctx context.Context) error {
+// WaitForNodeAge checks the node's age to see if its less than the minimum node age. If so, sleeps until the node
+// reaches the minimum node age.
+func WaitForNodeAge(client *kubernetes.Clientset, namespace string, minNodeAge time.Duration, ctx context.Context) error {
```
integrii, line 71: when supplying a context, it should always come first in the function signature
```suggestion
func WaitForNodeAge(ctx context.Context, client *kubernetes.Clientset, namespace string, minNodeAge time.Duration) error {
```

pkg/checks/external/nodeCheck/main.go
```diff
@@ -66,29 +66,34 @@ func WaitForKubeProxy(client *kubernetes.Clientset, namespace string, ctx contex
...
 
-	nodeNew, err := checkNodeNew(client, khPod, minNodeAge)
+	node, err := client.CoreV1().Nodes().Get(khPod.Spec.NodeName, v1.GetOptions{})
 	if err != nil {
 		return err
 	}
+	// get current age of the node
+	nodeAge := time.Now().Sub(node.CreationTimestamp.Time)
+	log.Debugln("Check running on node: ", node.Name, "with node age:", nodeAge)
+	if nodeAge < minNodeAge {
```
integrii, line 86:
```suggestion
	if nodeAge <= minNodeAge {
```

PR #548, opened 2020-07-07
Fix go routine leak and refactor concurrency on external checks

integrii:
Should be a more inclusive fix for #537


integrii:
Still not fixed in this branch.

image

integrii:
Not only is the memory leak gone, but memory use overall looks much lower after this PR...

image

pkg/checks/external/main.go
```diff
@@ -636,14 +613,16 @@ func (ext *Checker) RunOnce() error {
...
-	case <-timeoutChan:
+	case <-timeoutChan: // were out of time
 		ext.log("timed out waiting for pod to startup")
-		ext.cleanup()
 		return ext.newError("failed to see pod running within timeout")
-	case <-shutdownEventNotifyC:
-		ext.log("pod removed expectedly while waiting for pod to start running")
+	case err := <-podDeletedChan: // pod removed unexpectedly
+		if err != nil {
+			ext.log("error from pod shutdown watcher when when watching for checker pod to start:", err.Error)
```
jonnydawg, line 621: extra `when` in error message

pkg/checks/external/main.go
```diff
@@ -652,61 +631,54 @@ func (ext *Checker) RunOnce() error {
...
 		ext.log(errorMessage)
 		return ext.newError(errorMessage)
-	case <-shutdownEventNotifyC:
-		ext.log("got notification that pod has shutdown while waiting for it to report in")
-		hasUpdated, err := ext.doFinalUpdateCheck(lastReportTime)
+	case err := <-podDeletedChan: // pod was removed
 		if err != nil {
-			ext.log("got error when doing final check if pod has reported in after witnessing a pod removal", err)
-			return err
+			ext.log("error from pod shutdown watcher when when watching for checker pod to report results:", err.Error)
```
integrii, line 649:
```suggestion
			ext.log("error from pod shutdown watcher when watching for checker pod to report results:", err.Error)
```

pkg/checks/external/main.go
integrii, line 621:
```suggestion
			ext.log("error from pod shutdown watcher when watching for checker pod to start:", err.Error)
```

pkg/checks/external/main.go
```diff
@@ -407,123 +407,96 @@ func (ext *Checker) watchForCheckerPodShutdown(shutdownEventNotifyC chan struct{
...
+				if !ok {
+					ext.log("checker pod shutdown monitor saw a modified event and the object was not a pod. skipped.")
+					break
+				}
+				ext.log("checker pod shutdown monitor saw a modified event. the pod changed to ", p.Status.Phase)
+			case watch.Deleted: // we saw a deleted event, so notify upstream, but only once
+				if !haveNotified {
+					outChan <- nil
+					haveNotified = true
+				}
```
joshulyne, line 479: do we need to have a `haveNotified` or can we just break / return out of this forloop? It doesn't seem like we need to go through the rest of this forloop, until we reach a deleted event in which then we exit out of this function/watch?

pkg/checks/external/main.go
```diff
@@ -596,14 +571,15 @@ func (ext *Checker) RunOnce() error {
 		return err
 	}
 
-	// waiting for all checker pods are gone...
+	// always run a cleanup when done
+	defer ext.cleanup()
```
joshulyne, line 575: you have two `defer ext.cleanup()` in the same `RunOnce()` function in line 536,  i don't think it's needed since you specified this up top....?

pkg/checks/external/main.go
integrii: updated

pkg/checks/external/main.go
integrii: good catch! fixed and pushed

cmd/kuberhealthy/Makefile
```diff
@@ -1,12 +1,12 @@
-# IMAGE="kuberhealthy/kuberhealthy"
-# TAG="2.2.0"
+IMAGE="kuberhealthy/kuberhealthy"
+TAG="2.2.0"
```
jonnydawg, line 2: I think the tag should be `v2.2.2`

cmd/kuberhealthy/Makefile
```diff
@@ -1,12 +1,12 @@
...
 build:
-	docker build -t kuberhealthy/kuberhealthy:v2.2.2 -f Dockerfile ../../
-
+	docker build -t ${IMAGE}:${TAG} -f Dockerfile ../../
+build-dev:
+	docker build -t kuberhealthy:dev -f Dockerfile ../../
 push:
-	docker push kuberhealthy/kuberhealthy:v2.2.2
-
+	docker push kuberhealthy/${IMAGE}:${TAG}
```
jonnydawg, line 9: You don't need the extra `kuberhealthy/` here because it's included in your `IMAGE` var

pkg/checks/external/main.go
```diff
@@ -411,123 +411,94 @@ func (ext *Checker) watchForCheckerPodShutdown(shutdownEventNotifyC chan struct{
...
+		// watch for either an abort from upstream or removal of the selected pods
 		select {
-		case <-stoppedChan: // the watcher has stopped
-			// re-create a watcher and restart it if it closes for any reason
-			watcher = ext.startPodWatcher(listOptions, ctx)
-			go ext.waitForDeletedEvent(watcher.ResultChan(), sawRemovalChan, stoppedChan) // restart the watch
-		case <-sawRemovalChan: // we saw the watched pod remove
+		case <-ctx.Done(): // graceful shutdown signal
+			ext.log("pod shutdown monitor stopping gracefully")
+		case <-ext.waitForDeletedEvent(watcher): // we saw the watched pod remove
```
jonnydawg, line 428: it's possible that `waitForDeletedEvent(watcher)` can return an error to this channel when a `Watch.Error` event occurs. In that case, should the we recreate the `watcher` and call `waitForDeletedEvent(watcher)` again?

cmd/kuberhealthy/Makefile
integrii: nice catch

cmd/kuberhealthy/Makefile
integrii: done!

pkg/checks/external/main.go
integrii: I think it's too hard to implement retries here because we won't know if the error is because the supplied watcher is completely closed, or some other condition.  As long as we pass the error upstream, it's probably better to do the retry logic at a higher level in the code.

I have had bad experiences trying to make every thing that tries anything do retries...

Issue #547, opened 2020-07-02
Add nodeSelector option to deployment-check podSpec


zeleena:
**Add nodeSelector option to deployment-check podSpec**
Currently, the deployment-check will fail in a cluster with both Linux nodes and Windows nodes. The deployment-check creates a deployment with pods that run CHECK_IMAGE, a Linux image. Since the Linux image CHECK_IMAGE will schedule on Windows nodes, and there is no nodeSelector to prevent this, it will fail since Windows nodes can't run Linux containers.
Note: this is different but the same underlying problem as #546, except with the **deployment** check.

**Insufficient workarounds**
While it is true we could taint the Windows nodes and add a toleration, having the functionality to use the nodeSelector within the deployment-check's podSpec would be highly preferable so that this node taint does not affect other deployments in the cluster.

**Desired change**
This change would include adding a field for NodeSelector under the deployment's .spec.template.spec (a key within this podSpec block: https://github.com/Comcast/kuberhealthy/blob/v2.2.0/cmd/deployment-check/deployment.go#L88 )

To grab the nodeSelector key(s) and value(s), it should be configurable in the [values file](https://github.com/Comcast/kuberhealthy/blob/v2.2.0/deploy/helm/kuberhealthy/values.yaml). I was thinking it could either inherit the [current nodeSelector configuration of the primary deployment-check pod](https://github.com/Comcast/kuberhealthy/blob/v2.2.0/deploy/helm/kuberhealthy/values.yaml#L84), or be set with extra environment variables via the [extraEnvs key](https://github.com/Comcast/kuberhealthy/blob/v2.2.0/deploy/helm/kuberhealthy/values.yaml#L80).

**Why this is important**
This is an issue because kuberhealthy will report "OK": false for the deployment-check with error:
```
failed to create deployment within timeout
```
since Linux containers can't be ran on Windows nodes. Without the nodeSelector, I do not want to run the deployment-check (but would like to) since it currently reports misleading information about my cluster.

integrii:
Just like #546, we should do the following here:

- Add an environment variable to the helm chart for deployment check (maybe NODE_SELECTOR)
- Document the new environment variable in the README.md for the deployment check
- Read the environment variable into a global within the deployment check
- If the environment variable read is not blank (""), then set the node selector in the deployment spec that the pod creates
- Rev the deployment check version in both the makefile and helm chart

jonnydawg:
This has been addressed in a PR and is undergoing testing and debugging.

Issue #546, opened 2020-07-02
Add nodeSelector option to daemonset-check podSpec


zeleena:
**Add nodeSelector option to daemonset-check podSpec**
Currently, the daemonset-check will fail in a cluster with both Linux nodes and Windows nodes. The daemonset-check creates a daemonset with pods that run PAUSE_CONTAINER_IMAGE, a Linux image. Since the Linux image PAUSE_CONTAINER_IMAGE will schedule on Windows nodes, no there is no nodeSelector to prevent this, it will fail since Windows nodes can't run Linux containers.

**Insufficient workarounds**
While it is true we could taint the Windows nodes and add a toleration, having the functionality to use the nodeSelector within the daemonset-check's podSpec would be highly preferable so that this node taint does not affect other deployments in the cluster.

**Desired change**
This change would include adding a field for NodeSelector under the daemonset's .spec.template.spec (underneath this podSpec block: https://github.com/Comcast/kuberhealthy/blob/v2.2.0/cmd/daemonset-check/main.go#L261 )

To grab the nodeSelector key(s) and value(s), it should be configurable in the [values file](https://github.com/Comcast/kuberhealthy/blob/master/deploy/helm/kuberhealthy/values.yaml). I was thinking it could either inherit the [current nodeSelector configuration of the primary daemonset pod](https://github.com/Comcast/kuberhealthy/blob/v2.2.0/deploy/helm/kuberhealthy/values.yaml#L72), or be set with extra environment variables via the [extraEnvs key](https://github.com/Comcast/kuberhealthy/blob/v2.2.0/deploy/helm/kuberhealthy/values.yaml#L71).

**Why this is important**
This is an issue because kuberhealthy will report "OK": false for the daemonset-check with error "waiting for all pods to come online" since Linux containers can't be ran on Windows nodes. Without the nodeSelector, I do not want to run the daemonset-check since it currently reports misleading information about my cluster.

integrii:
Well described!  I think this should be pretty easy to add.  I think we should:

- Add an environment variable to the helm chart for daemonset check (maybe `NODE_SELECTOR`)
- Document the new environment variable in the `README.md` for the daemonset check
- Read the environment variable into a global within the daemonset check
- If the environment variable read is not blank (`""`), then set the node selector in the daemonset spec that the pod creates
- Rev the daemonset check version in both the makefile and helm chart

jonnydawg:
This has been addressed in a PR and is undergoing testing and debugging.

PR #549, opened 2020-07-07
add node selector to ds and deployment checks

jonnydawg:
Addresses #546 and #547

### DIFF
--- a/cmd/daemonset-check/README.md
+++ b/cmd/daemonset-check/README.md
@@ -10,9 +10,7 @@ does not complete within the given timeout it will report a timeout error on the
 
 Containers are deployed with their resource requirements set to 0 cores and 0 memory and use the pause container from
 Google (gcr.io/google_containers/pause:0.8.0), which is likely already cached on your nodes. The pause container is already used by kubelet to do various tasks and should be cached at all times. The node-role.kubernetes.io/master 
-NoSchedule taint is tolerated by daemonset testing pods. The Daemonset Check respects node selectors via the 
-`NODE_SELECTOR` environment variable to assign these daemonsets to specified nodes. If a failure occurs anywhere in
-the daemonset deployment or tear down, an error is shown on the status page describing the issue.
+NoSchedule taint is tolerated by daemonset testing pods. The Daemonset Check respects a comma separated list of `key=value` node selectors with the `NODE_SELECTOR` environment variable. If a failure occurs anywhere in the daemonset deployment or tear down, an error is shown on the status page describing the issue.
 
 #### Daemonset Check Kube Spec:
 
@@ -36,7 +34,9 @@ spec:
       - env:
           - name: POD_NAMESPACE
             value: "kuberhealthy"
-        image: kuberhealthy/daemonset-check:v3.0.0
+          - name: NODE_SELECTOR # Schedules daemonsets only to nodes with this label
+            value: "kubernetes.io/os=linux"
+        image: kuberhealthy/daemonset-check:v3.1.0
         imagePullPolicy: IfNotPresent
         name: main
         resources:
@@ -53,6 +53,7 @@ spec:
 |PAUSE_CONTAINER_IMAGE|"gcr.io/google-containers/pause:3.1"|
 |SHUTDOWN_GRACE_PERIOD|1m|
 |CHECK_DAEMONSET_NAME|"daemonset"|
+|NODE_SELECTOR|`<none>`|
 
 #### Daemonset Check Diagram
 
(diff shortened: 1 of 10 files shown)


cmd/daemonset-check/README.md
```diff
@@ -9,10 +9,10 @@ The check runs every 15 minutes (spec.runInterval), with a check timeout set to
 does not complete within the given timeout it will report a timeout error on the status page.
 
 Containers are deployed with their resource requirements set to 0 cores and 0 memory and use the pause container from
-Google (gcr.io/google_containers/pause:0.8.0), which is likely already cached on your nodes. The
-node-role.kubernetes.io/master NoSchedule taint is tolerated by daemonset testing pods. The pause container is already
-used by kubelet to do various tasks and should be cached at all times. If a failure occurs anywhere in the daemonset
-deployment or tear down, an error is shown on the status page describing the issue.
+Google (gcr.io/google_containers/pause:0.8.0), which is likely already cached on your nodes. The pause container is already used by kubelet to do various tasks and should be cached at all times. The node-role.kubernetes.io/master 
+NoSchedule taint is tolerated by daemonset testing pods. The Daemonset Check respects node selectors via the 
+`NODE_SELECTOR` environment variable to assign these daemonsets to specified nodes. If a failure occurs anywhere in
```
integrii, line 14: can you include examples here just like the deployment check has?

cmd/deployment-check/main.go
```diff
@@ -62,6 +62,10 @@ var (
 	checkDeploymentReplicasEnv = os.Getenv("CHECK_DEPLOYMENT_REPLICAS")
 	checkDeploymentReplicas    int
 
+	// Node selectors for the deployment check
+	checkDeploymentNodeSelectorsEnv = os.Getenv("NODE_SELECTOR")
+	checkDeploymentNodeSelectors    = make(map[string]string, 0)
```
integrii, line 67: no `0` required here

cmd/deployment-check/input.go
```diff
@@ -125,6 +125,18 @@ func parseInputValues() {
 		log.Infoln("Parsed CHECK_DEPLOYMENT_REPLICAS:", checkDeploymentReplicas)
 	}
 
+	// Parse incoming deployment node selectors
+	if len(checkDeploymentNodeSelectorsEnv) != 0 {
+		splitEnvVars := strings.Split(checkDeploymentNodeSelectorsEnv, ",")
+		for _, splitEnvVarKeyValuePair := range splitEnvVars {
+			parsedEnvVarKeyValuePair := strings.Split(splitEnvVarKeyValuePair, "=")
+			if _, ok := checkDeploymentNodeSelectors[parsedEnvVarKeyValuePair[0]]; !ok {
```
integrii, line 133: Why do we check for the key to exist here rather than just setting the key/value pair we want?

Also, I would add a couple len checks to ensure that the len of the things you're splitting is `2`.  If it's not, you continue and log an error because something isn't balanced.

integrii: whatever goes here also goes on the daemonset-check code...

jonnydawg: I first check for the key to exist incase a user inputs something similar to `foo=bar,foo=baz` -- in this case, should I check for a len of `3`? `a=x` has three chars because of the equals sign

cmd/deployment-check/input.go
```diff
@@ -125,6 +125,18 @@ func parseInputValues() {
...
+				checkDeploymentNodeSelectors[parsedEnvVarKeyValuePair[0]] = parsedEnvVarKeyValuePair[1]
+			}
+		}
+		log.Infoln("Parsed NODE_SELECTOR:", checkDeploymentNodeSelectors)
+	}
+
```
integrii, line 139:
```suggestion
	if len(checkDeploymentNodeSelectorsEnv) > 0 {
		splitEnvVars := strings.Split(checkDeploymentNodeSelectorsEnv, ",")
		for _, splitEnvVarKeyValuePair := range splitEnvVars {
			parsedEnvVarKeyValuePair := strings.Split(splitEnvVarKeyValuePair, "=")
			if len(parsedEnvVarKeyValuePair) != 2 {
			    log.Warningln("Unable to parse key value pair:",splitEnvVarKeyValuePair)
			    continue
			}
			checkDeploymentNodeSelectors[parsedEnvVarKeyValuePair[0]] = parsedEnvVarKeyValuePair[1]
			log.Infoln("Parsed NODE_SELECTOR:", parsedEnvVarKeyValuePair[0],parsedEnvVarKeyValuePair[1])
		}
	}

```

cmd/deployment-check/input.go
integrii: yeah, 3 for the length of string makes perfect sense there

PR #551, opened 2020-07-10
First cut at storage-check

ChrisHirsch:


integrii:
Nice!

Instead of us taking this into the main project, it would be acceptable to put this in your own repository (and push your own images somewhere), then have it linked in the [check registry](https://github.com/Comcast/kuberhealthy/blob/master/docs/EXTERNAL_CHECKS_REGISTRY.md).

This way you could release whenever you like and we would not be a blocker for any fixes or updates you want to push out.  I definitely want to make this work accessible via the project, but prefer not to stand in the way of its management and releases.  Does the strategy of self-hosting your check code sound acceptable to you?

ChrisHirsch:
This has been moved to ChrisHirsch/kuberhealthy-storage-check (and Gosec was fixed) and a new PR with updated documentation has been submitted.

kuzm1ch:
hi, btw, as a user, it will be better for me to have all checks in one place(especially for storage one). 
because:
1) it is better to rely on one docker hub repository 
https://github.com/Comcast/kuberhealthy/blob/master/deploy/helm/kuberhealthy/values.yaml#L146 
2) In case we will one place to store checks, they will have more attention in the main repo that in a separate one. 
https://github.com/ChrisHirsch/kuberhealthy-storage-check

if you can see this repo hasn't had enough attention from outside. Basically, if this code was in main repo the situation can be different as PVS availability is kind of basic of cluster health.
(especially when you on-premise env and rely on own storage provider)

ChrisHirsch:
I don't disagree, but I was asked to separate it out so that the core project can move forward and any additional plugins can maintain their own.

PR #553, opened 2020-07-13
Added external kuberhealthy storage check information

ChrisHirsch:
Documenting external storage check

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -2,3 +2,4 @@
 - [Jake Martin](mailto:jacob.martin@comcast.com)
 - [Ian Hoegen](mailto:ianhoegen@gmail.com)
 - [Babak Ghadiri](mailto:bbkghadiri6@gmail.com)
+- [Chris Hirsch](mailto:chris@base2technology.com)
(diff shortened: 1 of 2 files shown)


joshulyne:
This is awesome @ChrisHirsch! Thanks so much for contributing this!

I tried applying your storage check yaml configuration to test this locally, but ran into a few errors... In the configuration  https://github.com/ChrisHirsch/kuberhealthy-storage-check/blob/master/deploy/storage-check.yaml I believe you need a `-` before `image:` and after `containers:`
```
    containers:
    - image: chrishirsch/kuberhealthy-storage-check:v0.0.1
```
Once I fixed the configuration, I ran into an `ErrImagePull` error -- is this image available somewhere publicly?

Lastly, I noticed that in your install section https://github.com/ChrisHirsch/kuberhealthy-storage-check#install, the link in your install instructions `kubectl apply -f https://raw.githubusercontent.com/Comcast/kuberhealthy/2.0.0/cmd/storage-check/storage-check.yaml` is no longer available, I think you just need to update it to `https://github.com/ChrisHirsch/kuberhealthy-storage-check/blob/master/deploy/storage-check.yaml `!

ChrisHirsch:
@joshulyne thank you for reviewing my very apparent (now) documentation issues. Those have been fixed and I'm merging them to master shortly. 
As for the public image, it SHOULD be available here: 
https://hub.docker.com/r/chrishirsch/kuberhealthy-storage-check
docker pull chrishirsch/kuberhealthy-storage-check:v0.0.1

Can you let me know if it's not?

joshulyne:
@ChrisHirsch got your storage check running in our clusters! Thanks for the quick fixes! From what I understand with how this check runs (and please correct me if I'm getting it wrong!), but it seems like the check creates a pvc, then creates an init-job that mounts this pvc and initializes it with data, then a check-job is created that tries to mount this same pvc and I believe this is where I run into this error in our test clusters where this check-job fails to mount the volume with this error:
``` 
Warning  FailedAttachVolume  106s  attachdetach-controller  AttachVolume.Attach failed for volume "pvc-752d50e5-c6dd-11ea-b763-06d1500ebc7a" : Error attaching EBS volume "vol-0284349597b4cee9a" to instance "i-0cc4ba3699a4c5f8f": "InvalidVolume.ZoneMismatch: The volume 'vol-0284349597b4cee9a' is not in the same availability zone as instance 'i-0cc4ba3699a4c5f8f'\n\tstatus code: 400, request id: e94eec0f-5097-4648-9303-6a38663c1f47"
```

ChrisHirsch:
@joshulyne so you're apparently my first use case in EKS (I did this all on bare-metal). It appears that in your case, we create an EBS store in one AZ, claim it and then attempt to walk that store into other AZs when we check on each node. Does that sounds correct to you? If so, I'm definitely open to ideas on how to fix it.
My design was to create a simple PV that can be walked among all the desired nodes so that you know for sure the exact same storage is available everywhere. For a cloud, that may not be the best idea.

I can definitely stand up an EKS cluster to test with (or maybe see where I have one stuffed away) and attempt to reproduce and then determine a fix.

Thanks!

joshulyne:
@ChrisHirsch Ah that makes sense! Sounds good. We're gonna remove the 2.3.0 milestone for now as we're trying to make a new release asap but please let us know when the check is good to test again! We'd love to test it on our clusters, thanks so much!

PR #572, opened 2020-07-22
Http-content-check with new pkg nodeCheck

jdowni000:
Integrated the new pkg nodeCheck into the khcheck http-content-check.


cmd/http-content-check/Makefile
```diff
@@ -1,5 +1,6 @@
+  
 build:
-	docker build -t kuberhealthy/http-content-check:v1.3.2 -f Dockerfile ../../
+	docker build -t kuberhealthy/http-content-check:v1.4.2 -f Dockerfile ../../
```
joshulyne, line 3: maybe 1.4.0 instead of 1.4.2?

cmd/http-content-check/main.go
```diff
@@ -42,26 +44,53 @@ func main() {
 	if TargetString == "" {
 		reportErrorAndStop("No string provided in YAML")
 	}
+}
+
+func main() {
+
+	checkTimeLimit := time.Minute * 1
+	ctx, ctxCancel := context.WithTimeout(context.Background(), checkTimeLimit)
```
joshulyne, line 52: idt we need ctxCancel -- the context.WithTimeout cancels the context when it reaches `checkTimeLimit` already. so maybe `ctx, _ := context.WithTimeout(context.Background(), checkTimeLimit)` instead?

cmd/http-content-check/main.go
```diff
@@ -42,26 +44,53 @@ func main() {
...
+
+	kubernetesClient, err := kubeClient.Create("")
+	if err != nil {
+		log.Errorln("Error creating kubeClient with error" + err.Error())
+	}
+
+	err = nodeCheck.WaitForKuberhealthy(ctx)
+	if err != nil {
+		ctxCancel()
```
joshulyne, line 61: we can get rid of this from the suggestion above ^

cmd/http-content-check/main.go
```diff
@@ -42,26 +44,53 @@ func main() {
...
+		log.Errorln("Error waiting for kuberhealthy endpoint to be contactable by checker pod with error:" + err.Error())
+	}
+
+	err = nodeCheck.WaitForKubeProxy(ctx, kubernetesClient, "kuberhealthy", "kube-system")
+	if err != nil {
+		// ctxCancel()
```
joshulyne, line 67: remove ctxCancel!

cmd/http-content-check/Makefile
jdowni000: that makes sense I will adjust

cmd/http-content-check/main.go
jdowni000: agreed that is better. I have made adjustment

cmd/http-content-check/main.go
jdowni000: done

cmd/http-content-check/main.go
jdowni000: done

cmd/http-content-check/http-content-check.yaml
```diff
@@ -7,13 +7,13 @@ spec:
...
-      imagePullPolicy: IfNotPresent # uses local image if present
-      name: main
-      env:
-        - name: "TARGET_URL"
-          value: "http://httpbin.org/" # The URL that application will use to look for a specified string
-        - name: "TARGET_STRING"
-          value: "http" # The string that will be used to parse through provided URL
-        - name: "TIMEOUT_DURATION"
-          value: "30s" # Specifies the time limit for requests made by the client to the URL
+      - image: kuberhealthy/http-content-check:v1.4.2 # The image of the check you just pushed
```
joshulyne, line 10: 1.4.2 --> 1.4.0

jdowni000: accidently updated http-check yaml instead....I will fix with a push

Issue #554, opened 2020-07-13
Kuberhealthy gets stuck when a khcheck object is misconfigured


jonnydawg:
**Describe the bug**
The kuberhealthy master will run into an `InternalError` when looking for khc changes:

```
Error when watching for khcheck changes: InternalError
```

Related to #222 

**Steps To Reproduce**
Misconfigure any khcheck installed. An easy way to do this is to remove the `-` under `podSpec.containers`.

I think this issue is caused by the mismatching types here. Kuberhealthy expects `podSpec.containers` to be an array, but removing this `-` causes `podSpec.containers` to become and object instead of an array of objects.

**Expected behavior**
Misconfigure a check's `podSpec.containers` and observe that checks will no longer be scheduled. 

**Screenshots**
```
time="2020-07-13T18:11:10Z" level=error msg="Error listing check configuration resources khcheckcrd.KuberhealthyCheckList.Items: []khcheckcrd.KuberhealthyCheck: khcheckcrd.KuberhealthyCheck.Spec: khcheckcrd.CheckConfig.PodSpec: v1.PodSpec.Containers: []v1.Container: decode slice: expect [ or n, but found {, error found in #10 byte of ...|tainers\":{\"image\":\"d|..., bigger context ...|0123456789ab\"},\"spec\":{\"podSpec\":{\"containers\":{\"image\
```

**Versions**
 - Cluster OS: Debian
 - Kubernetes Version: 1.11.7
 - Kuberhealthy v2.2.0

jonnydawg:
Currently dealing with this issue from the webhook server:

```
2020/07/18 03:18:19 http: TLS handshake error from 192.168.65.3:54690: remote error: tls: bad certificate
2020/07/18 03:18:49 http: TLS handshake error from 192.168.65.3:54944: remote error: tls: bad certificate
```

PR #573, opened 2020-07-22
Fetch all khchecks using unstructured objects

joshulyne:
Addresses #554 

When listing all khchecks, if any of the khchecks are misconfigured, this functionality `khCheckClient.List(metav1.ListOptions{}, checkCRDResource, "")` fails to list any of the khchecks, blocking the main kuberhealthy process. To account for this potential error / misconfiguration of khchecks, using the dynamic client, fetch all khchecks as an unstructured object and proceed to convert each khcheck to the structured object `khcheck`.

For now, we skip / log the error if the a misconfigured khcheck cannot be converted to a structured object...


integrii:
@jonnydawg please review!

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -447,16 +463,19 @@ func (k *Kuberhealthy) addExternalChecks() error {
...
 	}
 
-	log.Debugln("Found", len(l.Items), "external checks to load")
+	log.Debugln("Found", len(khChecks.Items), "external checks to load")
 
 	// iterate on each check CRD resource and add it as a check
-	for _, r := range l.Items {
+	for _, kc := range khChecks.Items {
+		r, err := convertUnstructuredKhCheck(kc)
+		if err != nil {
```
integrii, line 476: Maybe add some logging here

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1215,3 +1234,45 @@ func (k *Kuberhealthy) configureInfluxForwarding() {
 	}
 	k.MetricForwarder = metricClient
 }
+
+func listUnstructuredKHChecks() (*unstructured.UnstructuredList, error){
+
+	var unstructuredList *unstructured.UnstructuredList
+	restConfig, err := clientcmd.BuildConfigFromFlags("", configPath)
+	if err != nil {
+		log.Errorln("Error building rest config for dynamic client:", err)
```
integrii, line 1243: I try not to log errors right where I return them.  I try to log errors where I decide what to do about them.  That would mean simply decorating this error and passing it upstream, then adding logging where you decide to `continue` past the entry.

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1215,3 +1234,45 @@ func (k *Kuberhealthy) configureInfluxForwarding() {
...
+		return unstructuredList, err
+	}
+
+	dynClient, err := dynamic.NewForConfig(restConfig)
+	if err != nil {
+		log.Errorln("Error creating dynamic client:", err)
+		return unstructuredList, err
```
integrii, line 1250: decorate error, remove logging in favor of single upstream log line.

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1215,3 +1234,45 @@ func (k *Kuberhealthy) configureInfluxForwarding() {
...
+
+	return unstructuredList, err
+}
+
+func convertUnstructuredKhCheck(unstructured unstructured.Unstructured) (khcheckcrd.KuberhealthyCheck, error){
+	un := unstructured.UnstructuredContent()
+	var khCheck khcheckcrd.KuberhealthyCheck
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(un, &khCheck)
+	if err != nil {
+		log.Errorln("Error converting unstructured object to khcheck:", err)
```
integrii, line 1273: remove logging, decorate error, rely on upstream code to log the message.

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -375,8 +387,12 @@ func (k *Kuberhealthy) monitorExternalChecks(ctx context.Context, notify chan st
 			}
 		}
 
-		// check for changes or additions in the incoming data
-		for _, i := range l.Items {
+		for _, kc := range khChecks.Items {
+			i, err := convertUnstructuredKhCheck(kc)
+			if err != nil {
```
integrii, line 392: add logging here because were deciding to do something based on an error that was returned

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -361,8 +367,14 @@ func (k *Kuberhealthy) monitorExternalChecks(ctx context.Context, notify chan st
 		// if a khcheck has been deleted, then we signal for change and purge it from the knownSettings map.
 		for mapName := range knownSettings {
 			var existsInItems bool // indicates the item exists in the item listing
-			for _, i := range l.Items {
-				itemMapName := i.Namespace + "/" + i.Name
+
+			for _, kc := range khChecks.Items {
+				khCheck, err := convertUnstructuredKhCheck(kc)
+				if err != nil {
+					continue
```
integrii, line 374: add logging here because were deciding to do something based on an error that was returned

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -348,11 +356,9 @@ func (k *Kuberhealthy) monitorExternalChecks(ctx context.Context, notify chan st
 		<-c
 		log.Debugln("Change notification received. Scanning for external check changes...")
 
-		// fetch all khcheck resources from all namespaces
-		l, err := khCheckClient.List(metav1.ListOptions{}, checkCRDResource, "")
+		khChecks, err := listUnstructuredKHChecks()
 		if err != nil {
-			log.Errorln("Error listing check configuration resources", err)
-			continue
+			log.Errorln("error listing unstructured khChecks: %w", err)
```
integrii, line 361: would it make sense to `continue` here?  Do we want to keep doing things in this iteration if we were unable to list unstructured checks?

joshulyne: yup, you're right, there should be a `continue` here

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1215,3 +1239,41 @@ func (k *Kuberhealthy) configureInfluxForwarding() {
...
+
+	return unstructuredList, err
+}
+
+func convertUnstructuredKhCheck(unstructured unstructured.Unstructured) (khcheckcrd.KuberhealthyCheck, error){
+	un := unstructured.UnstructuredContent()
+	var khCheck khcheckcrd.KuberhealthyCheck
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(un, &khCheck)
+	if err != nil {
+		return khCheck, errors.New("Error converting unstructured object to khcheck: " + err.Error())
```
integrii, line 1275: next time check out the new-ish `fmt.Errorf` function and use the `%w` variable to wrap the error in your new error... This way you can use the new `errors.Is` call to improve error handling.

integrii:
Let @jonnydawg review and then give 'er a mergin'.

PR #574, opened 2020-07-23
Deployment check patch for node check

jonnydawg:
Adds node check package to the deployment check -- implemented before the check is ran.


cmd/deployment-check/run_check.go
```diff
@@ -315,3 +320,25 @@ func cleanUpOrphanedResources(ctx context.Context) chan error {
...
+	err := nodeCheck.WaitForNodeAge(ctx, client, checkNamespace, time.Minute*2)
+	if err != nil {
+		log.Errorln("Failed to check node age:", err.Error())
+	}
+
+	// // Check if kube-proxy is ready.
+	// err = nodeCheck.WaitForKubeProxy(ctx, client, checkNamespace, "kube-system")
+	// if err != nil {
+	// 	log.Errorln("Failed to wait for kube-proxy to become ready:", err.Error())
+	// }
```
joshulyne, line 337: do we just want to remove this ^^?

jonnydawg:
I've built and pushed the image as well

PR #577, opened 2020-07-24
Add volumes to kuberhealthy deployment config

joshulyne:
I think we were missing the `volumes` for the configmap! @jdowni000


deploy/helm/kuberhealthy/templates/deployment.yaml
```diff
@@ -90,6 +90,10 @@ spec:
           requests:
             cpu: {{ .Values.resources.requests.cpu }}
             memory: {{ .Values.resources.requests.memory }}
+      volumes:
+      - configMap:
+          name: kuberhealthy
+        name: config-volume
```
integrii, line 96: I think this is already done up on line 31

PR #581, opened 2020-07-29
Create watcher for unstructured khchecks

joshulyne:
Fixes a bug where an error found from the watcher restarts the watcher every second...

### DIFF
--- a/cmd/kuberhealthy/kuberhealthy.go
+++ b/cmd/kuberhealthy/kuberhealthy.go
@@ -43,6 +43,12 @@ import (
 	"github.com/Comcast/kuberhealthy/v2/pkg/metrics"
 )
 
+// Set dynamicClient that represents the client used to watch and list unstructured khchecks
+var (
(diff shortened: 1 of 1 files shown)


cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1277,3 +1277,30 @@ func convertUnstructuredKhCheck(unstructured unstructured.Unstructured) (khcheck
...
+
+func watchUnstructuredKHChecks() (watch.Interface, error) {
+
+	var watcher watch.Interface
+	restConfig, err := clientcmd.BuildConfigFromFlags("", configPath)
+	if err != nil {
+		return watcher, err
+	}
+
+	dynClient, err := dynamic.NewForConfig(restConfig)
```
integrii, line 1289: Can we somehow re-use the single global Kubernetes client?  We had an issue long ago that stemmed from using multiple Kubernetes clients concurrently.

[this client](https://github.com/Comcast/kuberhealthy/blob/master/cmd/kuberhealthy/main.go#L78) is what I am proposing is used instead

Issue #582, opened 2020-07-29
Implement nodecheck waits on network connection check


integrii:
We should implement some code that ensures that the hosting node is ready to do network connections on the network connectivity check.  If the check runs too fast, then a node may not be able to connect to the target and that would cause a false alarm.

PR #583, opened 2020-07-31
Network check patch for node check

jonnydawg:
Added node check -- waits for node to be 5m old and that Kuberhealthy is reachable.

Also refactored to respect a `ctx` in some areas.

---

There was an issue where this check would sometimes get scheduled on a new node, before kube-proxy was available -- this attempts to address that.

### DIFF
--- a/cmd/network-connection-check/main.go
+++ b/cmd/network-connection-check/main.go
@@ -14,7 +14,9 @@
 package main
 
 import (
+	"context"
 	"errors"
+	"io/ioutil"
 	"net"
 	"os"
 	"strconv"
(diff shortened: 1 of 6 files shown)


jonnydawg:
Please take a look at this @bavarianbidi

jonnydawg:
#582

cmd/network-connection-check/main.go
```diff
@@ -84,7 +102,11 @@ func main() {
 	// create a new network connection checker
 	ncc := New()
 
-	err = ncc.Run(client)
+	ctx = context.Background()
```
joshulyne, line 105: can we do context with timeout for the nodeChecks?

PR #594, opened 2020-08-06
Add resources.limits Deployment template

zeleena:
Currently, the values file suggests that resources.limits can be set by these lines: https://github.com/Comcast/kuberhealthy/blob/master/deploy/helm/kuberhealthy/values.yaml#L21-L23

However, the current Deployment template does not ever use these resources.limits values, and only uses resources.requests as shown here: https://github.com/Comcast/kuberhealthy/blob/master/deploy/helm/kuberhealthy/templates/deployment.yaml#L89-L92

The Deployment template should be able to configure limits.cpu and limits.memory from the values file.

The other three static manifests that contain the same Deployment should likely also have these resources.limits values, since they do not have any limits as of now. I obtained the specific values from the the Helm chart's values.yaml, the first link above.

### DIFF
--- a/deploy/kuberhealthy-prometheus-operator.yaml
+++ b/deploy/kuberhealthy-prometheus-operator.yaml
@@ -423,6 +423,9 @@ spec:
           requests:
             cpu: 400m
             memory: 300Mi
+          limits:
+            cpu: 2
+            memory: 1Gi
       restartPolicy: Always
       terminationGracePeriodSeconds: 60
 ---
--- a/deploy/kuberhealthy-prometheus.yaml
+++ b/deploy/kuberhealthy-prometheus.yaml
@@ -426,6 +426,9 @@ spec:
           requests:
             cpu: 400m
             memory: 300Mi
+          limits:
+            cpu: 2
+            memory: 1Gi
       restartPolicy: Always
       terminationGracePeriodSeconds: 60
 ---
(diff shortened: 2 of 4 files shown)


deploy/kuberhealthy-prometheus-operator.yaml
```diff
@@ -423,6 +423,9 @@ spec:
           requests:
             cpu: 400m
             memory: 300Mi
+          limits:
```
jonnydawg, line 426: These files with get re-generated based on the helm chart -- doing this will cause it to be overwritten.

deploy/kuberhealthy-prometheus.yaml
```diff
@@ -426,6 +426,9 @@ spec:
           requests:
             cpu: 400m
             memory: 300Mi
+          limits:
```
jonnydawg, line 429: Same here

deploy/kuberhealthy-prometheus-operator.yaml
zeleena: I can remove them here and in deploy/kuberhealthy-prometheus.yaml. Is this also the case for deploy/kuberhealthy.yaml and should those lines be removed as well?

jonnydawg: Yes, this is also the case! I think this is okay as it shouldn't conflict with the result of the re-generation

PR #584, opened 2020-07-31
Add manual trigger feature for khchecks by creating kuberhealthy jobs crd

joshulyne:
#528 Manual trigger functionality using kuberhealthyjobs

PR #619, opened 2020-08-24
Reaper khjob

jdowni000:
Implemented a function to reap khjobs utilizing new function khJobDelete


integrii:
I think this one is blocked waiting on #584

integrii:
This should be unblocked after a change in #584

PR #623, opened 2020-08-27
fix: remove unused variable enableScraping

kuzm1ch:
* enableScraping is not used in helm template || should be delated
* update docs based on these changes
* additional minor change: make "With Prometheus Operator:" is under "Install kuberhealthy with Helm using one of the following..."


kuzm1ch:
@integrii

joshulyne:
@kuzm1ch thanks so much for catching this! lgtm! 👍

PR #627, opened 2020-08-31
Fix send report logic on Kubernetes 1.16 above

Bo0km4n:
## Why
In kubernetes 1.16 and above, a change was made to the routing process from Pod to Service, with remoteAddr being logged as the IP of Node.
This caused a mismatch in the request validation of the kuberhealthy server, resulting in a 400 error in reporting.
This fix solves the above problem by adding the IP of the pod to the report's json at the code level.

## What
- Added a member of PodIP to the Report structure used by khcheck-daemonset and khcheck-deployment
- Swap the order of json unmarshal processing and IP validation inside the process for request validation in kuberhealthy's server.
- Added POD_IP environment variable to khcheck-daemonset and khcheck-deployment PodSpecs


integrii:
Thanks for bringing this to the project's attention!  I would rather implement this fix in a slightly different way.  We have code that overrides environment variables on checker pods before they are created.  Adding the pod IP there would cause it to get added to all existing and future checks without any spec modifications.

It also looks like you may have accidentally enabled some checks in the helm chart by accident in this PR.

integrii:
Check this code [here](https://github.com/Comcast/kuberhealthy/blob/master/pkg/checks/external/main.go#L1091) and possibly open a fresh PR?

Bo0km4n:
@integrii  Thanks your replying and reviewing.
Am I correct in understanding that I can use the function `configureUserPodSpec` that overwrites the podSpec you provided to embed the pod's IP into the environment variables without changing the helm's chart or manifest?

One more question: is my PR okay with the kuberhealthy api server side of the fix?

Bo0km4n:
For now, I changed it to get the IP of the pod as an environment variable without changing the helm template.

Bo0km4n:
ping @integrii

integrii:
A few things here...

A) I am not sure that Kubernetes now reflects the node's IP from all pods.  Are you sure this wasn't just a part of your environment?  Sounds like a CNI configuration option...  We actually have a test that builds Kuberhelathy on Kubernetes 1.18 in a container and it passes with the code as-is.

B) There are several changes in this PR that are conflated.  I would prefer to do one thing at a time, even if that means opening several individual PRs.  
    - Changing default helm chart options in values.yaml
    - Adding some `nodecheck` package calls in the DNS check

C) The `Report` struct should not be modified to accomplish our goals here.

integrii:
Closing this one for now.  Please reopen if you circle back to this one.

PR #635, opened 2020-09-03
add more options for check-reaper cj

isaaguilar:
- added `startingDeadlineSeconds` to check-reaper's cronjob template

I've been running into issues where the check-reaper has too many missed schedules for whatever reason. I figured adding this is harmless and gives the user an option to configure this without having to patch manually.

### DIFF
--- a/deploy/helm/kuberhealthy/templates/check-reaper.yaml
+++ b/deploy/helm/kuberhealthy/templates/check-reaper.yaml
@@ -26,6 +26,9 @@ spec:
           restartPolicy: OnFailure
           serviceAccountName: check-reaper
   concurrencyPolicy: Forbid
+  {{- if .Values.checkReaper.startingDeadlineSeconds }}
+  startingDeadlineSeconds: {{ .Values.checkReaper.startingDeadlineSeconds }}
+  {{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
--- a/deploy/helm/kuberhealthy/values.yaml
+++ b/deploy/helm/kuberhealthy/values.yaml
@@ -128,3 +128,4 @@ checkReaper:
     repository: kuberhealthy/check-reaper
     tag: v1.4.0
   nodeSelector: {}
+  # startingDeadlineSeconds:
--- a/deploy/helm/kuberhealthy/Chart.yaml
+++ b/deploy/helm/kuberhealthy/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: kuberhealthy
 appVersion: 2.2.0
-version: 2.2.1
+version: 2.2.2
 home: https://comcast.github.io/kuberhealthy/
 description: "An operator for synthetic test and monitoring. Works great with Prometheus."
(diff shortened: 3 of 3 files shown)


deploy/helm/kuberhealthy/templates/check-reaper.yaml
```diff
@@ -26,6 +26,7 @@ spec:
           restartPolicy: OnFailure
           serviceAccountName: check-reaper
   concurrencyPolicy: Forbid
+  startingDeadlineSeconds: {{ .Values.checkReaper.startingDeadlineSeconds }}
```
integrii, line 29: Can we add `.Values.checkReaper.startingDeadlineSeconds` to the `values.yaml` file here?  That way we don't end up templating a blank property in the spec.

integrii (changes requested):
Thanks for adding this!  Makes sense to me.  Just one thing to review.

deploy/helm/kuberhealthy/templates/check-reaper.yaml
isaaguilar: I did add the item to `values.yaml` and left it intentionally blank since that it the default for a cronjob. There's no way to have a default value defined that would match the same behavior as a nil value. 

I was thinking of using an if-statement to only add the field when defined, but a blank spec achieves the same effect. If you think this might cause an issue rendering the template or cause undesired results, I can be a little more explicit. What do you think @integrii

joshulyne: @isaaguilar super sorry for the late reply! i believe that's the way we've been enabling helm values throughout the project if there are no default values provided so it'd be great to stay consistent that way!

isaaguilar: I added an if-statement around `startingDeadlineSeconds` so blank properties remain out of the spec. Hope this is what we were looking for.

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -128,3 +128,4 @@ checkReaper:
     repository: kuberhealthy/check-reaper
     tag: v1.4.0
   nodeSelector: {}
+  startingDeadlineSeconds: # Default is nil for CronJobs
```
joshulyne, line 131: similar to how we have it for deployments (https://github.com/Comcast/kuberhealthy/blob/058a6b3a3eb488547199809030aede878b13b632/deploy/helm/kuberhealthy/values.yaml#L38-L39) could you comment out this `startingDeadlineSeconds:`?

isaaguilar: Thanks for pointing this out. Forgot to update these values

joshulyne (approved):
lgtm! thanks so much for the quick turnaround!!!

joshulyne:
@integrii pinging you since you've gotta re-review!

integrii (approved):

bbhenry:
What's the plan to make these changes available in helm deployment? The only available versions in the index is 2.1.0 and 2.2.0. None of the minor versions are available for deployment with helm.

PR #643, opened 2020-09-14
Add AWS AMI Exists check to External Checks Registry

mtougeron:


joshulyne:
@mtougeron Thanks so much for contributing this! Tested this locally in our k8s cluster and it worked well!! @jonnydawg looking for your review since you contributed the original ami check but this looks GREAT to me!! 🎉

mtougeron:
> Tested this locally in our k8s cluster and it worked well!!

That's great! Hopefully this will be useful for folks who are not using kops

joshulyne (approved):

integrii:
This rocks!

integrii (approved):

mtougeron:
thanks!

PR #651, opened 2020-09-21
Ssl check minor fix

zjhans:
Fixes to check versions/names in spec sheets and readme files, comment typo.

### DIFF
--- a/cmd/ssl-expiry-check/main.go
+++ b/cmd/ssl-expiry-check/main.go
@@ -46,7 +46,7 @@ func init() {
 		return
 	}
 	insecureCheck = os.Getenv("INSECURE")
-	if len(daysToExpire) == 0 {
+	if len(insecureCheck) == 0 {
 		log.Error("ERROR: The INSECURE environment variable has not been set.")
 		return
 	}
--- a/cmd/ssl-expiry-check/main.go
+++ b/cmd/ssl-expiry-check/main.go
@@ -47,7 +47,7 @@ func init() {
 	}
 	insecureCheck = os.Getenv("INSECURE")
 	if len(daysToExpire) == 0 {
-		log.Error("ERROR: The DAYS environment variable has not been set.")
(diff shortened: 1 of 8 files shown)


cmd/ssl-expiry-check/main.go
```diff
@@ -47,7 +47,7 @@ func init() {
 	}
 	insecureCheck = os.Getenv("INSECURE")
 	if len(daysToExpire) == 0 {
```
integrii, line 49:
```suggestion
	if len(insecureCheck) == 0 {
```

cmd/ssl-selfsign-handshake-check/ssl-selfsign-handshake-check.yaml
```diff
@@ -15,7 +15,7 @@ spec:
           # If not using default SSL port of 443, port env variable must be updated  
           - name: PORT
             value: "443"
-        image: kuberhealthy/ssl-selfsign-handshake:v1.2.0
```
integrii, line 18: If you're changing image names, be sure that the github action (under `.github/workflows` at the root) has the right info in it.  I would also modify the `Makefile` for the check so that it builds with the right image name and tag.

integrii (changes requested):

cmd/ssl-expiry-check/main.go
zjhans: Fixed.

cmd/ssl-selfsign-handshake-check/ssl-selfsign-handshake-check.yaml
zjhans: Github actions under workflows should all match current image names of `_check-name_-check`.

All Makefiles match current check names, and are set to current versions, 1.2.0.

joshulyne (approved):
lgtm! @integrii final review!

integrii (approved):

PR #654, opened 2020-09-22
feat: add external dns status check to helm templates

kuzm1ch:
To reflect [external dns check yaml manifest](https://github.com/Comcast/kuberhealthy/blob/master/cmd/dns-resolution-check/externalDNSStatusCheck.yaml ) in helm templates


joshulyne:
@kuzm1ch Hmm I think the original reason to not add the external-dns-status check to our helm template was that some teams run air-gapped k8s clusters and so this check would fail for them... I know you've set `enabled: false` so this check wouldn't run automatically.... curious to get @integrii opinion.

integrii:
That was exactly it. We have to be very careful about what defaults we enable. I would rather see people have to opt-in to several things (and make that easy) than to risk Kuberhealthy being broken on some clusters.

integrii (approved):
This looks fine to me, without actually having tested all the options.  Perhaps we could include more environment variables (all of them) that the DNS check supports?

joshulyne:
> This looks fine to me, without actually having tested all the options. Perhaps we could include more environment variables (all of them) that the DNS check supports?

@kuzm1ch!

PR #658, opened 2020-09-24
Adds nodeChecks to SSL checks, fixes exits at end of check runs, adds…

zjhans:
… automatic self sign certificate pulling to handshake check


cmd/ssl-selfsign-handshake-check/main.go
```diff
@@ -42,23 +30,79 @@ func init() {
...
+		reportErr := reportKHSuccess()
+		if reportErr != nil {
+			log.Error(reportErr)
+		}
+		os.Exit(1)
+	}
+}
+
+// whichCheck determines if a cert has been provided, or if it needs to be retrieved from the host and returns a bool
+func whichCheck() bool {
```
joshulyne, line 74: maybe rename `whichCheck` to `certFileProvided` or something more specific to what this function is doing?

cmd/ssl-selfsign-handshake-check/main.go
```diff
@@ -42,23 +30,79 @@ func init() {
...
+	var err error
+
+	// if the user provided pem file is present, import it and run the handshake check
+	if fileFound == true {
+		err := ssl_util.HandshakeFromFile(domainName, portNum)
+		return err
+	}
+
+	// if the pem file has not been provided, pull the cert from the host, import it, and for the handshake check
+	if fileFound == false {
```
joshulyne, line 93:
```suggestion
	if !fileFound {
```

cmd/ssl-selfsign-handshake-check/main.go
```diff
@@ -42,23 +30,79 @@ func init() {
...
 	}
+	return fromFile
 }
 
-// run the SSL handshake check for the specified host and port number from ssl_util package
-func runSelfsignHandshake() error {
-	err := ssl_util.SelfsignHandshake(domainName, portNum)
+// runCheck takes in a bool from the whichCheck func and calls the one of two handshake check functions
+func runCheck(fileFound bool) error {
+	var err error
```
joshulyne, line 84: you can remove this and just `return nil` at the end

cmd/ssl-selfsign-handshake-check/main.go
```diff
@@ -42,23 +30,79 @@ func init() {
...
+		}
+
+		err = ssl_util.HandshakeFromHost(domainName, portNum, selfCert)
+
+		if err != nil {
+			log.Warn("Error performing handshake: ", err)
+		}
+		return err
+	}
 	return err
```
joshulyne, line 106: if you don't instantiate the `var err error` in the beginning you can just `return nil`

pkg/checks/external/ssl_util/main.go
```diff
@@ -17,37 +17,80 @@ import (
...
+
+	conn, err := tls.DialWithDialer(d, "tcp", host+":"+port, &tls.Config{
+		InsecureSkipVerify: true,
+		MinVersion:         tls.VersionTLS12,
+	})
+
+	if err != nil {
+		log.Warnln([]*x509.Certificate{&x509.Certificate{}}, "", err)
+		return hostCert, err
+	}
```
joshulyne, line 35: huh so there's potential here to return an empty cert...

pkg/checks/external/ssl_util/main.go
```diff
@@ -17,37 +17,80 @@ import (
...
 
+	// read the system cert pool, proceed with a new, empty pool if no certs found
+	rootCAs, _ := x509.SystemCertPool()
+	if rootCAs == nil {
+		rootCAs = x509.NewCertPool()
+	}
+
+	if cert != nil {
+		rootCAs.AddCert(cert)
+	}
```
joshulyne, line 65: and the cert gets checked here whether or not its empty... can we do that check outside of the function? like the check after you call the `CertificatePuller` to see if the result is empty.

joshulyne: what happens if it is empty?

pkg/checks/external/ssl_util/main.go
```diff
@@ -17,37 +17,80 @@ import (
...
 
 	err = conn.Handshake()
 	if err != nil {
 		log.Warn("Unable to complete SSL handshake with host ", host+": ", err)
 		return err
 	}
+
 	log.Info("SSL handshake to host ", host, " completed successfully")
 	return err
 }
```
joshulyne, line 88: seems like this is duplicated here:
https://github.com/Comcast/kuberhealthy/pull/658/files#diff-86d238c1d056bcbd1fb47e415fdb9ea8R115-R133

wondering if we can pull this out into a seperate function that makes the ssl handshake. Both the `HandshakeFromFile` and `HandshakeFromHost` pulls a cert that gets added to the `RootCAs`

joshulyne: seems like that function is right here in `CertHandshake`
https://github.com/Comcast/kuberhealthy/pull/658/files#diff-86d238c1d056bcbd1fb47e415fdb9ea8R137-R162

the only thing missing / different is the `RootCAs:            rootCAs,`

cmd/ssl-selfsign-handshake-check/main.go
zjhans: Updated func name.

cmd/ssl-selfsign-handshake-check/main.go
zjhans: Updated.

pkg/checks/external/ssl_util/main.go
zjhans: Updated all functions to throw errors if the self-signed certificates could not be pulled, or could not be appended to the certificate pool.

pkg/checks/external/ssl_util/main.go
zjhans: Updated to error out if an empty certificate is pulled from the host.

cmd/ssl-expiry-check/main.go
```diff
@@ -64,7 +92,9 @@ func main() {
 	reportErr := reportKHSuccess()
 	if reportErr != nil {
 		log.Error(reportErr)
+
 	}
+	os.Exit(1)
```
jonnydawg, line 97: If it reaches this point, it should be considered `successful` -- you should just exit with `0` here

cmd/ssl-expiry-check/README.md
```diff
@@ -22,15 +24,17 @@ spec:
       - env:
           # Domain name env variable must be updated to the domain on which you wish to check the SSL for
           - name: DOMAIN_NAME
-            value: "corporate.comcast.net"
+            value: "corporate.comcast.com"
```
jonnydawg, line 27: Maybe sanitize this? (possibly sensitive information)

cmd/ssl-expiry-check/main.go
```diff
@@ -50,10 +54,34 @@ func init() {
...
 		return
 	}
+	// set debug mode for nodeCheck pkg
+	nodeCheck.EnableDebugOutput()
 }
 
 func main() {
-	err := runExpiry()
+	// create context
+	checkTimeLimit := time.Minute * 1
```
jonnydawg, line 63: This works because the check runs quickly, but will bug out if the check takes longer than the `3m` `checkTimeout` that Kuberhealthy expects. Try using https://github.com/Comcast/kuberhealthy/blob/8889e96e6107513430ee533149cf6294b04b0119/pkg/checks/external/checkclient/main.go#L115 if you can

cmd/ssl-selfsign-handshake-check/README.md
```diff
@@ -49,51 +51,71 @@ metadata:
...
-    rjDMrEj4euJaL+K3G8cnMwH0T43NJQxdZ1pFCapa0zN8DKuzQBnqP2Kydn4Yu8HF
-    O2ghdlom1rdx0T3njhT00p5eYOkRLNM81yii4amufwgJjFXhu9tGC1bQAqamFA8/
-    eoLeEbuThio1NQ3WcWZVGGqDJ8d0rb91R08bt2Uj5ZmLm3MdbHaZJF5ACTgEo/Z8
-    L4xUYsxjVFj6q/QxqQ8yknpgJmtRkTAmwCJrwzgi3Gr47CEhZwHjx/BcHS5SXA8T
-    vYJXg5iBHvMnuRbfK0/+6iXdw4f/fsEpixi+w9TwvRgQGPsrBYHL+Ci3Hak6/+fA
-    DWXjT4yqkDhHvEXiQ/GV9nrEffE2c2ClW9s905mwPT7Ey09r4uzSWSu5ZtAOp3V1
-    ebvx36F3DOePQrZiLqei06YZjVFmEPlI1/DX0r0Xyw4JdMxxBzzYQOz/DuvfwxZ6
-    V4PGKNTve6HVj3Ar/aHL6DjvlPT5fyy970ZVbwXupHa2XJucCB+g8mMj/l1ubM4p
-    uio=
+    MIIDrDCCApSgAwIBAgIIdW/eznBkz2swDQYJKoZIhvcNAQELBQAwFTETMBEGA1UE
```
jonnydawg, line 54: Is this also sensitive info?

jonnydawg (changes requested):

cmd/ssl-selfsign-handshake-check/README.md
zjhans: This is a public SSL cert for a Docker Desktop Kubernetes local cluster. The parsed cert info is as follows:
Certificate Information:
Common Name: kube-apiserver
Subject Alternative Names: docker-for-desktop, kubernetes, kubernetes.default, kubernetes.default.svc, kubernetes.default.svc.cluster.local, kubernetes.docker.internal, vm.docker.internal, localhost, IP Address:10.96.0.1, IP Address:0.0.0.0, IP Address:192.168.65.3, IP Address:127.0.0.1
Valid From: January 15, 2020
Valid To: January 14, 2021
Serial Number: 8462227203082735467 (0x756fdece7064cf6b)

I think it's safe.

cmd/ssl-expiry-check/README.md
zjhans: This is our public facing, corporate site. I chose that as opposed to "google.com" because Google seems to renew their certs very frequently and keep them constantly on the verge of expiring, and the default check length of 60 days fails..

cmd/ssl-expiry-check/main.go
zjhans: Ack, updated for next commit.

cmd/ssl-expiry-check/main.go
```diff
@@ -15,19 +15,31 @@
...
+var (
+	domainName    string
+	portNum       string
+	daysToExpire  string
+	insecureCheck string
+	insecureBool  bool
+)
+
+// CheckTimeout is a variable for how long code should run before it should retry.
+var CheckTimeout time.Duration
```
joshulyne, line 42: i don't think you need to make this a global var -- seems like its only used in `init()`?

cmd/ssl-expiry-check/main.go
```diff
@@ -64,11 +115,13 @@ func main() {
 	reportErr := reportKHSuccess()
 	if reportErr != nil {
 		log.Error(reportErr)
+
 	}
+	os.Exit(0)
 }
 
+// runExpiry runs the SSL expiry check from the ssl_util package with the specified env variables
```
joshulyne, line 123: in this func:
`if certExpired == true` --> `if certExpired`
`if expirePending == true` --> `if expirePending`

pkg/checks/external/ssl_util/main.go
```diff
@@ -17,67 +19,148 @@ import (
 
 var TimeoutSeconds = 10
 
-// CertHandshake performs a basic TLS/SSL handshake on the specified host/port, returning any errors
-func CertHandshake(host, port string) error {
-	log.Info("Testing SSL handshake on host ", host, " over port ", port)
+// FileUploaded returns 'true' if the pem formatted cert has been mounted at /etc/ssl/selfsign/certificate.crt
+func FileUploaded() bool {
```
joshulyne, line 23: some of these functions don't look like they're used outside of this file -- you can lowercase them `fileUploaded()`, `CertFromHost()`, `PoolFromHost()`, etc.

cmd/ssl-expiry-check/main.go
zjhans: This is for the enhancement in the comment above, for using the `GetDeadline` func. Removed until I have that working.

pkg/checks/external/ssl_util/main.go
zjhans: Lowercased all functions not used outside `ssl_util` package.

cmd/ssl-expiry-check/main.go
zjhans: Updated.

cmd/ssl-expiry-check/main.go
zjhans: GetDeadline functions for timeout checking added to both checks.

cmd/ssl-expiry-check/main.go
```diff
@@ -50,37 +87,103 @@ func init() {
...
+
+	// set debug mode for nodeCheck pkg
+	nodeCheck.EnableDebugOutput()
 }
 
 func main() {
-	err := runExpiry()
+	// create context
+	checkTimeLimit := time.Minute * 1
+	ctx, _ := context.WithTimeout(context.Background(), checkTimeLimit)
```
joshulyne, line 100: so this ctx is for the nodeCheck -- maybe we can rename it to that? since its confusing with the ctx you reset below for the actual check run.

also maybe rename `checkTimeLimit` to `nodeCheckTimeout`?

cmd/ssl-expiry-check/main.go
```diff
@@ -109,3 +212,17 @@ func reportKHFailure(errorMessage string) error {
...
+	err := nodeCheck.WaitForNodeAge(ctx, client, *namespace, time.Minute*5)
+	if err != nil {
+		log.Errorln("Failed to check node age:", err.Error())
+	}
+
+	// Check if Kuberhealthy is reachable.
+	err = nodeCheck.WaitForKuberhealthy(ctx)
+	if err != nil {
+		log.Errorln("Failed to reach Kuberhealthy:", err.Error())
+	}
```
joshulyne, line 227: this is repeated: https://github.com/Comcast/kuberhealthy/pull/658/files#diff-d7ac0c34789683f7b6664eede5c96959R108-R112

you already run this check `WaitForKuberhealthy` in your main func

cmd/ssl-handshake-check/main.go
```diff
@@ -15,17 +15,52 @@
...
+
+type Checker struct {
+	client         *kubernetes.Clientset
+	domainName     string
+	portNum        string
+	selfSignedBool bool
+}
 
 func init() {
+	var err error
```
joshulyne, line 51: is this needed? seems like its instantiated in line 57

cmd/ssl-handshake-check/main.go
```diff
@@ -36,26 +71,92 @@ func init() {
...
+
+	// set debug mode for nodeCheck pkg
+	nodeCheck.EnableDebugOutput()
 }
 
 func main() {
-	err := runHandshake()
+	// create context
+	checkTimeLimit := time.Minute * 1
+	ctx, _ := context.WithTimeout(context.Background(), checkTimeLimit)
```
joshulyne, line 88: same comment here as in https://github.com/Comcast/kuberhealthy/pull/658/files#r498511466

joshulyne (commented):
a few minor comments, but overall lgtm! this looks awesome! 👍

cmd/ssl-handshake-check/main.go
zjhans: Updated.

cmd/ssl-expiry-check/main.go
zjhans: Removing from `main` and keeping in `waitForNodeToJoin`

pkg/checks/external/ssl_util/main.go
zjhans: I believe this has been updated satisfactorily in the last PR. Resolving.

cmd/ssl-handshake-check/main.go
zjhans: Updated, PR resubmitted.

cmd/ssl-expiry-check/main.go
zjhans: Updated, PR resubmitted.

joshulyne (approved):
lgtm! 👍

jonnydawg (approved):

Issue #634, opened 2020-09-02
kuberhealthy/deployment error from /metrics have a wrong string format


darleilopes:
**Describe the bug**
The check `kuberhealthy/deployment` error is not closing correctly the string at `error` value, were Prometheus can't consuming this metric cause it cannot run parse.
 
**Steps To Reproduce**
Invoking `kuberhealthy/deployment` check producing an error.

**Expected behavior**
Change the metric output `error="deployments.apps is forbidden: User "system:serviceaccount:kuberhealthy:default" cannot create resource "deployments" in API group "apps" in the namespace "kuberhealthy"|"` to, for example `error="deployments.apps is forbidden: User 'system:serviceaccount:kuberhealthy:default' cannot create resource 'deployments' in API group 'apps' in the namespace 'kuberhealthy'|"`, or another format that allow close the string correctly.

**Screenshots**
The `/metrics` endpoint is returning the following output:
```bash
kuberhealthy_check{check="kuberhealthy/deployment",namespace="kuberhealthy",status="0",error="deployments.apps is forbidden: User "system:serviceaccount:kuberhealthy:default" cannot create resource "deployments" in API group "apps" in the namespace "kuberhealthy"|"} 0
```
**Versions**
 - Cluster OS: Debian 10
 - Kubernetes Version: 1.16.6
 - Kuberhealthy Version: 2.2.0

integrii:
Looks like our error metadata needs sanitized so that the prometheus metrics don't become unparsable.  Thanks for bringing this up!  We will look at sanitizing the error messages in our Prometheus metric data.

darleilopes:
One thing I could see is that when this error occurred, my prometheus stopped seeing all metrics, in which I see a risk of losing control of other events due to this error

PR #660, opened 2020-09-24
fixing the metrics format for the error messages

brahminikatta:
#634 fixing the metrics format

### DIFF
--- a/pkg/metrics/exporter.go
+++ b/pkg/metrics/exporter.go
@@ -14,6 +14,7 @@ package metrics // import "github.com/Comcast/kuberhealthy/v2/pkg/metrics"
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/metrics/exporter.go
```diff
@@ -50,6 +51,7 @@ func GenerateMetrics(state health.State) string {
 				errors += fmt.Sprintf("%s|", error)
 			}
 		}
+		errors = strings.Replace(errors, "\"", "", -1)
```
jonnydawg, line 54: Try using the `strings.ReplaceAll()` func here -- and be careful, I think we want to replace the quotes `"` with single quotes `'` instead of removing them.

jonnydawg (changes requested):

jonnydawg (approved):
Looks good!

PR #665, opened 2020-09-30
Add Support for PodSecurityPolicy

### LINKED ISSUE #664: The helm chart does not provide a PodSecurityPolicy
The helm chart is already supporting a SecurityContext. But it is lacking a matching PodSecurityPolicy.
A PodSecuritPolicy and the matching RBAC should ease deployment on kubernetes clusters with enabled PodSecurityPolicy.

czunker:
Fixes #664

### DIFF
--- a/deploy/helm/kuberhealthy/templates/podsecuritypolicy.yaml
+++ b/deploy/helm/kuberhealthy/templates/podsecuritypolicy.yaml
@@ -1,0 +1,53 @@
+{{- if .Values.podSecurityPolicy.enabled }}
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: runtime/default
+    seccomp.security.alpha.kubernetes.io/defaultProfileName: runtime/default
+  name: {{ template "kuberhealthy.name" . }}-psp
+  labels:
+    app: {{ template "kuberhealthy.name" . }}
+spec:
+  privileged: false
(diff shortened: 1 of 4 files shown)


deploy/helm/kuberhealthy/templates/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,53 @@
+{{- if .Values.podSecurityPolicy.enabled }}
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: runtime/default
+    seccomp.security.alpha.kubernetes.io/defaultProfileName: runtime/default
+  name: {{ template "kuberhealthy.name" . }}-psp
```
integrii, line 9: We might not need `-psp` here because the `kind` is already set to `PodSecurityPolicy`.

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -60,6 +60,11 @@ securityContext:
   allowPrivilegeEscalation: false
   readOnlyRootFilesystem: true
 
+# When enabled kuberhealthy will create a PodSecurityPolicy, Role and Binding to match the specified securityContext
+# In case you use your own images for tests, you might need to add other PSPs as well on your own
+podSecurityPolicy:
+  enabled: false
```
integrii, line 66: Thank you for defaulting this to false.

integrii (approved):
This looks great!  Thank you for taking the extra steps to add this and comment it well.

PR #671, opened 2020-10-06
Add error text from fetchPodByIP to log

DenisPalnitsky:
`fetchPodByIP` function could fail due to multiple reasons but the reason for failure is not reflected in the log which makes it hard to troubleshoot


joshulyne (approved):
that makes sense! thank you!!

Issue #650, opened 2020-09-18
Create new cronjob run check


integrii:
Cronjobs can sometimes get stuck and miss 100 runs.  When they do this, Kubernetes will stop scheduling them entirely.  We should be alerted when cronjobs have stopped.

- catch when a cron job has skipped some number of runs
    - either by looking at last run time, or
    - by looking for an event that says 100 events have been missed
- allow an optional whitelist environment variable that restricts which cronjobs in a namespace are analyzed

jdowni000:
PR has been submitted for this. #677

PR #677, opened 2020-10-09
Cronjob checker

jdowni000:
The cronjob-checker reaches out to a specified cronjob in a given namespace and lists events. It then ranges over the events for a specified reason. If there is an event with the specified reason in the given cronjob it will alert kuberhealthy. Issue #650


cmd/cronjob-checker/main.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	//range over eventList for cronjob events that match provided reason
+	for _, e := range eventList.Items {
+		// log.Debugln(e.Reason)
+		if reason == e.Reason {
+			log.Infoln("There was an event with reason:" + e.Reason + " for cronjob" + cronJob)
+			reportErr := fmt.Errorf("cronjob: " + cronJob + " has an event with reason: " + reason)
+			ReportFailureAndExit(reportErr)
+		}
```
joshulyne, line 78: only question here, is how do we know the failure is coming from the cronjob? not sure if the cronjob name is being searched anywhere in the events, only the reason?

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1151,6 +1151,7 @@ func (k *Kuberhealthy) healthCheckHandler(w http.ResponseWriter, r *http.Request
 // this will return the state of ALL found checks.
 // Failures to fetch CRD state return an error.
 func (k *Kuberhealthy) getCurrentState(namespaces []string) health.State {
+	log.Debugln("getCurrentState: ", namespaces)
```
joshulyne, line 1154: is this intentional?

deploy/helm/kuberhealthy/templates/configmap.yaml
```diff
@@ -28,6 +28,14 @@ data:
...
         annotations:
           description: Kuberhealthy shows that the cluster is not healthy
+      - alert: DeadMansSwitch
+        expr: vector(1)
+        labels:
+          severity: critical
+        annotations:
+          description: This is a DeadMansSwitch meant to ensure that the entire Alerting
+          pipeline is functional.
+        summary: Alerting DeadMansSwitch
```
joshulyne, line 38: curious what this is for?

joshulyne:
just a few comments! wondering if we can expand this to check on all cronjobs in a given namespace maybe? instead of just checking one cronjob at a time?

cmd/kuberhealthy/kuberhealthy.go
jdowni000: this was not...perhaps Eric and I wrote this when we were debugging something earlier...have no idea how that got in there

deploy/helm/kuberhealthy/templates/configmap.yaml
jdowni000: these must be coming from merging...although I thought this was only master that I had been doing it too....I am not sure but will get rid of it

integrii:
- Include the cronjob name that is failing in the kuberhealthy error report
- remove the reason env var and hard-code it into place (for now)
- consider removing namespace env var reference in the readme, because it is automatically set to the namespace the check gets deployed into

cmd/cronjob-checker/README.md
```diff
@@ -0,0 +1,88 @@
...
+  runInterval: 1h
+  timeout: 10m
+  podSpec:
+    serviceAccountName: cronjob-checker
+    containers:
+      - name: cronjob-checker
+        image: kuberhealthy/cronjob-checker:v1.1.0
+        imagePullPolicy: IfNotPresent
+        env:
+          - name: NAMESPACE
```
integrii, line 36: True this up with the `fieldRef` in your example yaml.

integrii (changes requested):

cmd/cronjob-checker/README.md
jdowni000: done

integrii:
- remove all namespaces from flag spec files in readme
- add specs to helm chart so that this deploys into the helm-specified namespace
- change clusterrolebindings to rolebindings

cmd/cronjob-checker/main.go
joshulyne: if theres one event failure, it reports failure and the program exits... what if there are multiple cronjob failures? i think we should be able to catch them all 😏

cmd/cronjob-checker/README.md
```diff
@@ -0,0 +1,78 @@
...
+      - "events.k8s.io"
+    resources:
+      - events
+    verbs:
+      - get
+      - list
+      - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
joshulyne, line 55: the example uses `ClusterRoleBinding` while the `cmd/cronjob-checker.yaml` uses `RoleBinding`

cmd/cronjob-checker/cronjob-checker.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - "events.k8s.io"
+    resources:
+      - events
+    verbs:
+      - get
+      - list
+      - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
```
joshulyne, line 44: should this be `ClusterRoleBinding`?

cmd/cronjob-checker/README.md
jdowni000: I have moved this to a Role and Role Binding

cmd/cronjob-checker/cronjob-checker.yaml
jdowni000: I have moved this to a Role and Role Binding

go.sum
```diff
@@ -392,15 +392,23 @@ honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc/go.mod h1:rf3lG4BRIbNafJWh
 honnef.co/go/tools v0.0.1-2019.2.3/go.mod h1:a3bituU0lyd329TUQxRnasdCoJDkEUEAqEt0JzvZhAg=
 k8s.io/api v0.19.0 h1:XyrFIJqTYZJ2DU7FBE/bSPz7b1HvbVBuBf07oeo6eTc=
 k8s.io/api v0.19.0/go.mod h1:I1K45XlvTrDjmj5LoM5LuP/KYrhWbjUKT/SoPG0qTjw=
+<<<<<<< HEAD
```
integrii, line 395: remove these lines to resolve the conflict markers pls

integrii (changes requested):

go.sum
jdowni000: I have removed 392-394

integrii (changes requested):
just a few small changes - otherwise this LGTM.  We can circle around and make `findLastCronRunTime` more reasonable later.

cmd/cronjob-checker/README.md
```diff
@@ -0,0 +1,58 @@
...
+  name: cronjob-checker
+spec:
+  runInterval: 5m
+  timeout: 10m
+  podSpec:
+    serviceAccountName: cronjob-checker
+    containers:
+      - name: cronjob-checker
+        image: kuberhealthy/cronjob-checker:v2.0.0
+        imagePullPolicy: Always
```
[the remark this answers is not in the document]
jdowni000: done

cmd/cronjob-checker/cronjob-checker.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: cronjob-checker
+spec:
+  runInterval: 5m
+  timeout: 10m
+  podSpec:
+    serviceAccountName: cronjob-checker
+    containers:
+      - name: cronjob-checker
+        image: kuberhealthy/cronjob-checker:v2.0.0
+        imagePullPolicy: Always
```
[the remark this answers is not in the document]
jdowni000: done

cmd/cronjob-checker/main.go
```diff
@@ -22,16 +22,25 @@ var kubeConfigFile = filepath.Join(os.Getenv("HOME"), ".kube", "config")
...
+		log.Fatalln("Error parsing KH_CHECK_RUN_DEADLINE:", err)
+	}
+	khCheckRunDeadline := time.Unix(intkhCheckRundDeadline, 0)
+
+	khCheckRunTime := khCheckRunDeadline.Sub(time.Now()) / 2
 
-	// create context
-	checkTimeLimit := time.Minute * 1
-	ctx, _ := context.WithTimeout(context.Background(), checkTimeLimit)
+	log.Infoln(khCheckRunDeadline)
```
integrii, line 42: For production code, let's either drop this log line.  Also, leaving a comment on the check run deadline code above would be nice.

cmd/cronjob-checker/main.go
```diff
@@ -81,12 +90,11 @@ func main() {
 		lastRunTimeV1 := cronGet.Status.LastScheduleTime
 		lastRunTime := lastRunTimeV1.Time
 		shouldOfRun := findLastCronRunTime(schedule)
-		timeMinus := shouldOfRun.Add(time.Duration(-30) * time.Second)
-		timePlus := shouldOfRun.Add(time.Duration(30) * time.Second)
+		tM, tP := scheduleWindow(shouldOfRun, time.Minute*5)
```
integrii, line 93: `tM` and `tP` are a little ambiguous here.  Could we rename this to something like `earliestRunTime` and `latestRunTime`

cmd/cronjob-checker/main.go
```diff
@@ -81,12 +90,11 @@ func main() {
...
 
 		log.Infoln("last run time for cronjob", c.Name, "was", lastRunTime)
 
-		if lastRunTime.Before(timeMinus) && lastRunTime.Before(timePlus) {
+		if lastRunTime.Before(tM) && lastRunTime.Before(tP) {
```
integrii, line 97:
```suggestion
		if lastRunTime.After(tM) && lastRunTime.Before(tP) {
```

What we want to say here is: "if the last run time was after the `earliestAllowedTime` and the last run time was before the `latestAllowedTime`.

integrii (changes requested):

cmd/cronjob-checker/main.go
jdowni000: removed and comment added

cmd/cronjob-checker/main.go
jdowni000: done

cmd/cronjob-checker/main.go
jdowni000: done. changed logic to make this the positive

integrii (approved):
Looks great!  Thanks!

Issue #640, opened 2020-09-08
Implement context.TODO() entries


integrii:
During an upgrade of the Kubernetes package versions in PR #636 a bunch of `context.TODO()` function call parameters were added.  We should go through and plug these into any existing contexts we have and implement new contexts for the others.

jonnydawg:
This has been done for deployment & daemonset

integrii:
This was done for the main binary.  Closing!

PR #676, opened 2020-10-08
clean up context.TODO()s and bump version tag

jonnydawg:
Finishes up the remaining `context.TODO()s` in the deployment check.

Also bumps the version tag to `v1.6.1` (which includes #669 and these changes)

PR #678, opened 2020-10-10
deployment-check: propagate context from entrypoint

hypnoglow:
Part of https://github.com/Comcast/kuberhealthy/issues/640

Since the context is now propagated all the way down from the `main()`, I'm not sure if we really need package variables `ctx` and `ctxCancel`. Should I remove these package variables?


hypnoglow:
Oops, probably a duplicate of https://github.com/Comcast/kuberhealthy/pull/676

jonnydawg:
Thanks for catching the ones I missed; I will do a quick testing on this!

jonnydawg (approved):

PR #682, opened 2020-10-14
Helm repo gh actions auto update

chadbitzer:
workflow and shell script have been written to update helm repo files and Chart.yaml with either the publishing of a release tag ("appVersion:" and "version:"), or a push to "./deploy/helm/kuberhealthy/*", except for Chart.yaml, which is auto-updated by the shell script just prior to the packaging process.


chadbitzer:
I've tested this with a positive release tag match, release tag non-matches, no tag push to "deploy/helm/kuberhealthy", and branch name starting with "v", and all have worked as expected.  This uses $GITHUB_RUN_NUMBER to track iterations of packages, which will increment by one each time this particular workflow runs.

joshulyne (approved):

PR #691, opened 2020-10-20
feat(helm): add template for external storage check

kuzm1ch:
Consider external storage check https://github.com/ChrisHirsch/kuberhealthy-storage-check from @ChrisHirsch very useful as it covers a critical component of cluster health(storage).

Reimplement https://github.com/ChrisHirsch/kuberhealthy-storage-check/tree/master/deploy manifests into helm template:
* range function allow testing several storage class with one template (DRY), by default, it will test default storage class

* use rootless images( as default one is alpine with root user)
```
     extraEnvs:
      CHECK_STORAGE_IMAGE: bitnami/nginx:1.19
      CHECK_STORAGE_INIT_IMAGE: bitnami/nginx:1.19 
```


joshulyne:
sorry for the late review!! i think this just needs to resolve with whats in master and it should be good to go! 👍   thanks so much!! @kuzm1ch

joshulyne (approved):

integrii (approved):
This is off by default so I think its a safe/helpful addition.  Thank you very much for augmenting the helm chart! LGTM.  This will be linted before going to the helm repo as well, so that adds an additional level of "it should work" here.

PR #692, opened 2020-10-20
adding logic to reap khjobs to check-reaper

jdowni000:
I have merged over the files from my original branch reaper-khjob. I have built a new image with this and am running it in lab successfully. Please review docs and let me know if there is anything else you would like added

### DIFF
--- a/cmd/check-reaper/main.go
+++ b/cmd/check-reaper/main.go
@@ -221,15 +221,6 @@ func deletePod(ctx context.Context, client *kubernetes.Clientset, pod v1.Pod) er
 	return client.CoreV1().Pods(pod.Namespace).Delete(ctx, pod.Name, options)
 }
 
-// podConditions returns true if conditions are met to be deleted for checker pod
-func podConditions(pod v1.Pod, duration float64, phase v1.PodPhase) bool {
-	if time.Now().Sub(pod.CreationTimestamp.Time).Hours() > duration && pod.Status.Phase == phase {
-		log.Infoln("Found pod older than ", duration, " hours in status ", phase, " .Deleting pod:", pod)
-		return true
-	}
-	return false
-}
-
 // jobConditions returns true if conditions are met to be deleted for khjob
 func jobConditions(job khjobcrd.KuberhealthyJob, duration time.Duration, phase khjobcrd.JobPhase) bool {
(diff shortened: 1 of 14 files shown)


cmd/check-reaper/main.go
```diff
@@ -195,3 +220,58 @@ func deletePod(ctx context.Context, client *kubernetes.Clientset, pod v1.Pod) er
 	options := metav1.DeleteOptions{PropagationPolicy: &propagationForeground}
 	return client.CoreV1().Pods(pod.Namespace).Delete(ctx, pod.Name, options)
 }
+
+// podConditions returns true if conditions are met to be deleted for checker pod
+func podConditions(pod v1.Pod, duration float64, phase v1.PodPhase) bool {
```
joshulyne, line 225: is this func being used anywhere?

cmd/check-reaper/main.go
```diff
@@ -195,3 +220,58 @@ func deletePod(ctx context.Context, client *kubernetes.Clientset, pod v1.Pod) er
...
+}
+
+// KHJobDelete fetches a list of khjobs in a namespace and will delete them if they meet given criteria
+func khJobDelete(client *khjobcrd.KHJobV1Client) error {
+
+	opts := metav1.ListOptions{}
+	del := metav1.DeleteOptions{}
+
+	// convert JobDeleteMinutes into time.Duration
+	JobDeleteTimeDuration, err := time.ParseDuration(JobDeleteTimeDurationEnv)
```
joshulyne, line 249: nitpick!
```suggestion
	jobDeleteTimeDuration, err := time.ParseDuration(JobDeleteTimeDurationEnv)
```

joshulyne (commented):
few nitpicky comments but lgtm!

cmd/check-reaper/main.go
jdowni000: Does not look like it...I am not sure why this function is saying it is an addition though this is not code that I wrote...I moved it to near the bottom though and added func jobConditions right below it....but yes this can be deleted

cmd/check-reaper/main.go
jdowni000: good catch I will make this not a global

joshulyne (approved):
lgtm!

PR #700, opened 2020-10-22
Daemonset service account name

### LINKED ISSUE #684: Assigning serviceaccount to deamonset check
Is that something thats a available? went through the code, couldnt find anything about it

integrii:
The flat spec files here were an accidental change from my local testing, but they will get regenerated after this merge anyway.  This should fix #684


jonnydawg (approved):
Thanks!

erezo9:
Thanks for the PR!

Issue #186, opened 2019-07-20
Implement node functionality checker that deploys a pod to every available node in the cluster


integrii:
Cordoned nodes are used when the daemonset checker's daemonset is deployed.  We should always ignore cordoned nodes because they are likely known to be bad.

We should modify the daemonset checker spec to have a node anti-affinity policy (or otherwise) to achieve this behavior.

rukatm:
From https://kubernetes.io/docs/concepts/architecture/nodes/

> Note: Pods created by a DaemonSet controller bypass the Kubernetes scheduler and do not respect the unschedulable attribute on a node. This assumes that daemons belong on the machine even if it is being drained of applications while it prepares for a reboot.

integrii:
Whoa.  Thanks for finding this.  I wonder if we need to move the checker to _not_ be a daemonset in order to get the functionality we want...

joshulyne:
https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#taints-and-tolerations
^^ this lists out the tolerations automatically added to the daemonset and it includes the `node.kubernetes.io/unschedulable:NoSchedule` toleration. 

Instead of rewriting the daemonset checker, we should consider writing a new khcheck that verifies pods are scheduled on everything except cordoned nodes.

integrii:
looks like we can't prevent this.  Changing to a new issue for adding a new node functionality check

PR #701, opened 2020-10-27
Ns pod checker

jdowni000:
namespace pod check retrieves a list of namespaces in a cluster and deploys a test pod in each one. If this is successful, it will delete the pod and report to kuberhealthy success if all namespaces are able to be deployed and deleted in. Otherwise, it will report the failure. This is in response to #186


integrii (approved):

integrii:
It would be nice to have some unit tests for some of the pieces here.  You can use a [fake kube client](https://godoc.org/k8s.io/client-go/kubernetes/fake) to do this.

PR #702, opened 2020-10-27
Adds CPU and memory request parameters to check-reaper cronjob

zjhans:
Jobs will fail to schedule on some clusters if CPU and memory requests are not explicitly set in spec.


jdowni000 (approved):
lgtm

PR #718, opened 2020-10-30
Fix Prometheus rules being ignored and broken helm charts for Prometheus

linkvt:
Hi,

this pull request fixes two issues:
- the previous ConfigMap with Prometheus Rules is not picked up as it is deprecated since 2018-08, see https://github.com/prometheus-operator/prometheus-operator/releases/tag/v0.23.0
- The whitespace between both ConfigMaps was chomped causing the ConfigMaps to merge, see https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace
  - I see that whitespace chomping is used quite extensively, I think this might need to be done a bit more sensitively but this is just my opinion

I would really appreciate a quick new release of the helm chart since master (and the published versions) are currently broken for Prometheus users, see above the `apiVersion` line executed on the `master` branch:
```
$ helm template --set prometheus.enabled=true  --set prometheus.enableAlerting=true deploy/helm/kuberhealthy 2>/dev/null | grep -C4 "not healthy---"
        for: 5m
        labels:
          severity: critical
        annotations:
          description: Kuberhealthy shows that the cluster is not healthy---
apiVersion: v1
kind: ConfigMap
metadata:
  name: kuberhealthy
``` 

Thanks a lot!


integrii (approved):
Thank you!  We really appreciate you fixing this. 👍

integrii:
You can expect the Helm chart to publish next time we do a new tagged release.  We should be able to do one shortly (v2.3.1).

Issue #715, opened 2020-10-28
KH_EXTERNAL_REPORTING_URL not respected


davidkarlsen:
**Describe the bug**

When overriding KH_EXTERNAL_REPORTING_URL through env-var it is not used by kh:

`k logs kuberhealthy-57f6557465-dsvhb |grep svc`:
```
time="2020-10-28T00:04:53Z" level=info msg="External check reporting URL set to: http://kuberhealthy.kuberhealthy.svc.cluster.local/externalCheckStatus"
```

`k get pod kuberhealthy-57f6557465-dsvhb -o yaml`:
```
...
spec:
  automountServiceAccountToken: true
  containers:
  - args:
    - -dsPauseContainerImageOverride
    - fsnexus.evry.com:8085/google_containers/pause:3.1
    command:
    - /app/kuberhealthy
    env:
    - name: POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: POD_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
    - name: KH_EXTERNAL_REPORTING_URL
      value: http://kuberhealthy.kuberhealthy.svc.icp-global.local/externalCheckStatus
    image: fsnexus.evry.com:8085/kuberhealthy/kuberhealthy:v2.3.0
    imagePullPolicy: IfNotPresent
    livenessProbe:
      failureThreshold: 3
```

**Steps To Reproduce**
Install chart v32 with values:
```
    image:
      repository: fsnexus.evry.com:8085/kuberhealthy/kuberhealthy
    prometheus:
      enabled: true
    deployment:
      env:
        KH_EXTERNAL_REPORTING_URL: http://kuberhealthy.kuberhealthy.svc.icp-global.local/externalCheckStatus
      nodeSelector:
        fs.evry.com/finods-group: aot
      args:
      - -dsPauseContainerImageOverride
      - fsnexus.evry.com:8085/google_containers/pause:3.1
    podSecurityPolicy:
      enabled: true
    check:
      daemonset:
        image:
          repository: fsnexus.evry.com:8085/kuberhealthy/daemonset-check
        extraEnvs:
          PAUSE_CONTAINER_IMAGE: "fsnexus.evry.com:8085/google-containers/pause:3.1"
      deployment:
        image:
          repository: fsnexus.evry.com:8085/kuberhealthy/deployment-check
        extraEnvs:
[...]

PR #719, opened 2020-11-04
Fixes variable assignment to allow use of custom external reporting U…

zjhans:
…RLs - issue #715

### DIFF
--- a/cmd/kuberhealthy/main.go
+++ b/cmd/kuberhealthy/main.go
@@ -105,8 +105,9 @@ func init() {
 	// set env variables into config if specified
 	externalCheckURL, err := getEnvVar(KHExternalReportingURL)
 	if err != nil {
-		cfg.ExternalCheckReportingURL = externalCheckURL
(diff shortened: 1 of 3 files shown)


cmd/kuberhealthy/main.go
```diff
@@ -105,8 +105,9 @@ func init() {
 	// set env variables into config if specified
 	externalCheckURL, err := getEnvVar(KHExternalReportingURL)
 	if err != nil {
-		cfg.ExternalCheckReportingURL = externalCheckURL
+		log.Infoln("KH_EXTERNAL_REPORTING_URL environment variable not set: ", err, "Attempting to use default value")
 	}
+	cfg.ExternalCheckReportingURL = externalCheckURL
```
joshulyne, line 110: it looks like we're doing this validation twice. Once here, and then another time in lines 130-136. I'd rather try and combine them if possible..

joshulyne: If the env var is not set `externalCheckURL` would be empty... and then go through the second validation below, which sort of seems repetitive?

joshulyne (approved):
lgtm!

Issue #685, opened 2020-10-15
Avoid using generic labels in checks [daemonset-check]


randrusiak:
Hi folks!
I started testing kuberhealthy in this week and I like it a lot. 
I have got a small problem with daemonset-check because it using "app" label with ephemeral names of created pods and every day it creates multiple entries in my logging system (grafana loki). 

![image](https://user-images.githubusercontent.com/29524175/96133632-4446c180-0efb-11eb-968f-21f1ef8c0dc9.png)

I think we should avoid using generic labels and instead of use labels for example with a prefix such as "kh_". I've looked at the source code and this label is using only for getting proper pod to delete so renaming shouldn't cause any problems.

joshulyne:
Hey @randrusiak thanks for the feedback! This totally makes sense to change in the daemonset check. Feel free to put up a pr yourself if you'd like to do so :)

PR #723, opened 2020-11-06
Updates label for daemonset checks from 'app' to less generic 'kh', f…

zjhans:
…ix for #685


cmd/daemonset-check/main.go
```diff
@@ -236,7 +236,7 @@ func reportOKToKuberhealthy() {
 func reportToKuberhealthy(ok bool, errs []string) {
 	var err error
 	if ok {
-		err = kh.ReportSuccess()
+		err := kh.ReportSuccess()
```
joshulyne, line 239: this isn't needed? can just be `=`

joshulyne (approved):
lgtm! thanks for taking on all these issues 👍

PR #731, opened 2020-11-20
Helm: ServiceMonitor Endpoints specifications

geoL86:
Helm: ServiceMonitor Endpoints specifications should be parametrised

### DIFF
--- a/.ci/values.yaml
+++ b/.ci/values.yaml
@@ -6,7 +6,12 @@ prometheus:
   enabled: false
   name: "prometheus"
   release: prometheus-operator
-  serviceMonitor: false
+  serviceMonitor:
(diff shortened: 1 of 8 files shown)


joshulyne (approved):
lgtm! thanks so much for taking this up!

joshulyne:
ah theres a `.ci/values.yaml` that we use for our e2e tests that also needs the modification: 
```
prometheus:
  enabled: false
  name: "prometheus"
  release: prometheus-operator
  serviceMonitor: false
  enableAlerting: false
```

joshulyne (changes requested):
just one more place to add the helm change!

joshulyne (approved):

PR #741, opened 2020-11-25
Adds toleration support to KH checks.

zjhans:


.ci/values.yaml
```diff
@@ -65,6 +70,11 @@ check:
       tag: v3.2.1
     extraEnvs:
     nodeSelector: {}
+    tolerations:
```
joshulyne, line 73: maybe put `tolerations: []` so we know it takes in a list?

joshulyne (approved):
other than the nitpicky comment, lgtm!

.ci/values.yaml
zjhans: Updated, now with brackets! 😃

integrii:
As soon as the comments in the `.ci/values.yaml` are pruned, this looks good to go!

joshulyne (approved):

Issue #752, opened 2020-12-09
Remove waits for kube-proxy


integrii:
Waiting for kube-proxy to be ready in our checks makes the assumption that the user has `kube-proxy`.  In newer clusters running Cilium (for example), kube-proxy does not exist.

We should remove this check from all of our checks so that it runs for everyone.  This was brought to attention by @bavarianbidi on our #kuberhealthy slack.

![image](https://user-images.githubusercontent.com/98695/101670260-ca691b00-3a07-11eb-8819-dca30c78e990.png)

Issue #739, opened 2020-11-25
[helm] DNS checks are missing priviledges


czunker:
**Describe the bug**
The internal and external DNS checks try to query the k8s API, but are missing privileges.
The check completes, but it seems it wants to do more.

```
$ k -n kuberhealthy logs dns-status-internal-1606308364
...
time="2020-11-25T12:46:11Z" level=debug msg="Getting pod: dns-status-internal-1606308364 in order to get its node information"
time="2020-11-25T12:46:11Z" level=error msg="Error waiting for node to reach minimum age: pods \"dns-status-internal-1606308364\" is forbidden: User \"system:serviceaccount:kuberhealthy:default\" cannot get resource \"pods\" in API group \"\" in the namespace \"kuberhealthy\""
...
time="2020-11-25T12:46:11Z" level=debug msg="Getting pod: dns-status-internal-1606308364 in order to get its node information"
time="2020-11-25T12:46:11Z" level=error msg="Error waiting for kube proxy to be ready: error getting kuberhealthy pod: pods \"dns-status-internal-1606308364\" is forbidden: User \"system:serviceaccount:kuberhealthy:default\" cannot get resource \"pods\" in API group \"\" in the namespace \"kuberhealthy\""
...
```

**Steps To Reproduce**
- Deploy kuberhealthy with enabled internal and external DNS checks

**Expected behavior**
No error messages in the log.

**Screenshots**

**Versions**
 - Cluster OS: Ubuntu 20.04
 - Kubernetes Version: 1.18.8
 - Kuberhealthy Release or build 2.3.1

**Additional context**

I'm not quite sure whether the API access is needed or not.

czunker:
Perhaps, a possible solution is, to add a specific service account, like the one for the daemonset check. This SA could get the needed privileges to query the API.

TBeijen:
Noticing the same thing. Tests seem to pass though.

Looks like it's introduced in #519, where updating helm chart with needed RBAC is overlooked.

integrii:
Thanks for reporting this.

It looks like this is caused by our `nodecheck` package trying to auto-detect which node the checker pod lives on in order to be sure the node is old enough to run the check properly (sometimes there is a race condition right as nodes start up).

[...]

PR #755, opened 2020-12-10
Remove nodecheck WaitForKubeProxy and modify WaitForNodeAge

joshulyne:
Removes nodecheck `WaitForKubeProxy` and modifies `WaitForNodeAge` so that you pass in the nodeName using the downward API instead of needing extra permissions to get the kuberhealthy pod in order to get nodeName information.

Ups all check versions with the nodeCheck updates.

Addresses #739 and #752


integrii (approved):
Awesome.  Should satisfy #752

PR #757, opened 2020-12-10
Adds additional labels option for ServiceMonitor

### LINKED ISSUE #742: ServiceMonitor should allow specifying custom labels so that prometheus-operator can discover them
Currently ServiceMonitor template has fixed set of labels and doesn't allow specifying additional labels. If we allow specifying that, then existing prometheus-operator that are configured to discover ServiceMonitors based on some label will work fine. As of now, we have to disable ServiceMonitor template and deploy custom one with the matching label(which our prometheus-operator uses). 

**Additional context**

**additional_labels** should be a map of additional labels, which can be inserted in template ar at runtime.

Something on these lines, 
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  labels:
    app: {{ template "kuberhealthy.name" . }}
    prometheus: {{ .Values.prometheus.name }}
    release: {{ .Values.prometheus.release }}
    {{ .Values.prometheus.serviceMonitor.additional_labels }}
  name: {{ template "kuberhealthy.name" . }}
  namespace: {{ .Release.Namespace }}
```

zjhans:
fixes #742

### DIFF
--- a/deploy/helm/kuberhealthy/templates/servicemonitor.yaml
+++ b/deploy/helm/kuberhealthy/templates/servicemonitor.yaml
@@ -8,6 +8,9 @@ metadata:
     app: {{ template "kuberhealthy.name" . }}
     prometheus: {{ .Values.prometheus.name }}
     release: {{ .Values.prometheus.release }}
+    {{- if .Values.prometheus.serviceMonitor.additional_labels}}
+    additional_labels: {{ .Values.prometheus.serviceMonitor.additional_labels}}
+    {{- end  -}}
   name: {{ template "kuberhealthy.name" . }}
   namespace: {{ .Release.Namespace }}
 spec:
--- a/deploy/helm/kuberhealthy/templates/servicemonitor.yaml
+++ b/deploy/helm/kuberhealthy/templates/servicemonitor.yaml
@@ -8,8 +8,8 @@ metadata:
     app: {{ template "kuberhealthy.name" . }}
     prometheus: {{ .Values.prometheus.name }}
     release: {{ .Values.prometheus.release }}
-    {{- if .Values.prometheus.serviceMonitor.additional_labels}}
-    additional_labels: {{ .Values.prometheus.serviceMonitor.additional_labels}}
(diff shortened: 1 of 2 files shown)


deploy/helm/kuberhealthy/templates/servicemonitor.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
     app: {{ template "kuberhealthy.name" . }}
     prometheus: {{ .Values.prometheus.name }}
     release: {{ .Values.prometheus.release }}
+    {{- if .Values.prometheus.serviceMonitor.additional_labels}}
+    additional_labels: {{ .Values.prometheus.serviceMonitor.additional_labels}}
+    {{- end  -}}
```
joshulyne, line 13: can we change it to use camelCase? I believe thats part of the helm best practices: https://helm.sh/docs/chart_best_practices/values/#naming-conventions

zjhans: Done!

deploy/helm/kuberhealthy/templates/servicemonitor.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
     app: {{ template "kuberhealthy.name" . }}
     prometheus: {{ .Values.prometheus.name }}
     release: {{ .Values.prometheus.release }}
+    {{- if .Values.prometheus.serviceMonitor.additionalLabels}}
+    additionalLabels: {{ .Values.prometheus.serviceMonitor.additionalLabels}}
```
joshulyne, line 12: sorry just thought of this, but with list values, i think there needs to be some extra configuration to make sure you get add every individual item... something like:
```
        {{- if .Values.deployment.args }}
        args:
{{ toYaml .Values.deployment.args | nindent 8 }}
        {{- end }}
```
https://github.com/Comcast/kuberhealthy/blob/master/deploy/helm/kuberhealthy/templates/deployment.yaml#L44-L47

joshulyne (approved):

PR #758, opened 2020-12-11
fix: add helper method to create CheckClient

jstrachan:
so that we can create a CheckClient if we already have a rest.Interface instance for connecting to k8s clients

this is a similar method to the CreateClient we have in the khstatecrd package


integrii (approved):
LGTM.  No danger in this change.  Thanks!

PR #759, opened 2020-12-16
Set default values for MaxPodsThreshold and JobDeleteTimeDuration

adriananeci:
In https://github.com/Comcast/kuberhealthy/pull/692, 2 new environment variables were added for check-reaper: `MAX_PODS_THRESHOLD` and `JOB_DELETE_TIME_DURATION`. If any of these environment variables are unset, then we'll get errors like:
```
level=error msg="Error converting MaxPodsThreshold to int"
```
or
```
time="2020-12-16T13:51:14Z" level=error msg="Error converting JobDeleteTimeDurationEnv to Float"
time="2020-12-16T13:51:14Z" level=error msg="Failed to reap khjobs with error:  time: invalid duration "
```

This PR sets default values in case these environment variables are unset or their values cannot be parsed.

### DIFF
--- a/cmd/check-reaper/main.go
+++ b/cmd/check-reaper/main.go
@@ -70,7 +70,7 @@ func init() {
 	// Parse incoming jobDeleteTimeDurationEnv environment variable.
 	JobDeleteTimeDuration = JobDeleteTimeDurationDefault
 	if len(JobDeleteTimeDurationEnv) != 0 {
-		// convert JobDeleteMinutes into time.Duration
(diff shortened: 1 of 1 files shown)


joshulyne:
This makes sense. Thanks for making all the improvements in the log messages as well! Before, we had set values and then made it more configurable.. Adding those default values makes it less error prone and help it run more smoothly! Thanks for pointing it out and adding your fixes!

joshulyne (approved):
lgtm!

PR #763, opened 2020-12-17
Allow overriding resources for checks #760

### LINKED ISSUE #760: Add resource requests and limits for checks to helm chart
Most if not all Kuberhealthy checks (present under templates directory) are currently using hard-coded resource requests and no limits at all. Same is true for 'check-reaper' too. It will be great **if we are able to provide/override resources** (the cpu/memory requests/limits) from values.yaml (by a map of resources)

**Additional context**
In our Kubernetes environment, we are mandating specifying the resource requests/limits as a matter of best practices. We also run terratest to detect the pods that are missing those. So if I want to adhere to that, I will need to maintain local version of those check definitions so as to override the resource limits from there. 
It will be more intuitive to expose those via values.yaml

AshutoshNirkhe:
Allow to override resources from values.yaml instead of hard-coded values. Also this will help users specify the resource limits for cpu/memory. 
For now I have just updated it for daemonset check. If reviewers like the idea, I can update all the check definitions in this PR! This will help resolve #760

### DIFF
--- a/deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
@@ -26,9 +26,7 @@ spec:
 {{- end }}
       resources:
         requests:
-          {{- if .Values.check.networkConnection.resources.requests.cpu }}
           cpu: {{ .Values.check.networkConnection.resources.requests.cpu }}
-          {{- end }}
           memory: {{ .Values.check.networkConnection.resources.requests.memory }}
         {{- if .Values.check.networkConnection.resources.limits }}
         limits:
--- a/deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
@@ -26,7 +26,19 @@ spec:
 {{- end }}
       resources:
         requests:
-          memory: 5Mi
+          {{- if .Values.check.networkConnection.resources.requests.cpu }}
+          cpu: {{ .Values.check.networkConnection.resources.requests.cpu }}
+          {{- end }}
+          memory: {{ .Values.check.networkConnection.resources.requests.memory }}
+        {{- if .Values.check.networkConnection.resources.limits }}
+        limits:
+          {{- if .Values.check.networkConnection.resources.limits.cpu }}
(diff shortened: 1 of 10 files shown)


integrii:
This looks good so far to me!  This should support the case that there are no limits specified as well as individual limits, but not both.

AshutoshNirkhe:
> This looks good so far to me! This should support the case that there are no limits specified as well as individual limits, but not both.

Yeah exactly, so I will be checking for presence of resources.limits first and then individual limits for cpu and memory as done for daemonset check. I will proceed with other checks changes then. Thanks!

AshutoshNirkhe:
@integrii added 'resources' for all checks. Please have a look. Thanks!

deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
```diff
@@ -26,7 +26,19 @@ spec:
 {{- end }}
       resources:
         requests:
-          memory: 5Mi
+          {{- if .Values.check.networkConnection.resources.requests.cpu }}
+          cpu: {{ .Values.check.networkConnection.resources.requests.cpu }}
+          {{- end }}
```
joshulyne, line 31: this is the only resource.requests that get checked if set or not? seems like we always want to have requests values required but not necessarily limits so we don't need this if check?

AshutoshNirkhe: @joshulyne to be honest, I looked into existing hard-coded values for resource requests/limits and added the conditions as per that. It turns out that this is the only check without cpu requests as in https://github.com/Comcast/kuberhealthy/blob/master/deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml#L28 not sure why. Hence I added that condition only for this check.

joshulyne: i agree with you, this is something we should have required when we added this check. i'm thinking setting `cpu: 10m` would be okay. would you mind adding that to this pr and removing the if statement?

joshulyne (approved):
thanks!! lgtm 👍

deploy/helm/kuberhealthy/templates/khcheck-network-connection.yaml
AshutoshNirkhe: Sure thing , updated :)

Issue #761, opened 2020-12-17
Default checks fail with segmentation violation


annkam:
Good afternoon! I have installed kuberhealthy v2.3.1 via Helm chart to my OKD 3.11 (kubernetes v1.11) cluster and have enabled default daemonset, deployment and dns-internal checks. They all seem to be running OK but fail to report to the main kuberhealthy pod. The error message is not really informative so I can't figure out the problem. 

daemonset-check is v3.2.1
deployment-check is v1.6.2
dns-resolution-check is v1.4.2

daemonset:

![daemonset](https://user-images.githubusercontent.com/18022078/102454223-3f73ca80-4070-11eb-8682-a55d1c941f55.png)

deployment:

![deployment](https://user-images.githubusercontent.com/18022078/102454254-4f8baa00-4070-11eb-9846-00efd4848bfb.png)

dns-status-internal:

![dns-status-internal](https://user-images.githubusercontent.com/18022078/102454285-5b776c00-4070-11eb-8bac-11843961974a.png)

joshulyne:
thanks for pointing this out -- definitely a bug in the checkClient pkg under `pkg/checks/external/checkClient`. Looking at that error, it seems like its actually on line 99 where its just writing a debug log message: 
https://github.com/Comcast/kuberhealthy/blob/master/pkg/checks/external/checkclient/main.go#L99
and I think it runs into this panic when we try and call `err.Error()` in the log message. i'll submit a pr for this!

integrii:
Relevant to #765

integrii:
This should be fixed in #765 This should be fixed now from PR #765, but the checks you are using probably haven't had the fixed check client package compiled into them yet...

Leaving this open because we need to rebuild all checks and up their tags.

PR #765, opened 2020-12-29
Remove err.Error() from writeLog message

joshulyne:
Addresses #761 

Possible fix for bug when sending report to kuberhealthy -- calling err.Error() in `writeLog` might not be valid, so removing and just using `err` similar to how we have it for all the other writeLog calls. 

I believe this error could have started when I made a change to add exponential backoff retry to the sendReport function


integrii:
This will get switched up to two `if` statements to avoid mixing up the situation of a bad return code and an error.

joshulyne:
@integrii if you look right below that code block, we already do a status check.... we didn't even need the OR D:

integrii (approved):

PR #771, opened 2021-01-05
Fix 'no Auth Provider found for name azure'

adriananeci:
While testing an external check outside an k8s cluster using oidc authentication, I'm getting 
```
no Auth Provider found for name "azure"
```
This PR attempt to fix such issues when validating checks outside a k8s cluster that uses external sources for obtaining credentials.

More details can be found at https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/client-go/examples

### DIFF
--- a/pkg/kubeClient/client.go
+++ b/pkg/kubeClient/client.go
@@ -13,6 +13,7 @@ package kubeClient // import "github.com/Comcast/kuberhealthy/v2/pkg/kubeClient"
 
 import (
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


integrii (approved):
Sounds like a great addition! Thanks for picking the package that adds support for all providers and thank you for opening a PR here.

integrii:
This change won't take effect into all of our checks until they are rebuilt, so we might need to kick off builds and up versions if anyone reports another issue like this.

PR #773, opened 2021-01-07
build should not 'error' when skipping master builds

integrii:
We have decided that we only want helm charts to be available after we tag a new release, but the current build script for the helm chart "errors" out with `exit 1` instead of `exit 0`.  This change allows the github action to complete successfully as expected.


rjacks161 (approved):
lgtm

PR #781, opened 2021-01-18
added: 100 as the default PASSING_PERCENT for http-check

### LINKED ISSUE #780: http-check: not specifying PASSING_PERCENT throws error
**Describe the bug**
http-check results in `"Error converting PASSING_PERCENT to int: strconv.Atoi: parsing \"\": invalid syntax"` when PASSING_PERCENT is not provided as an env var.

**Steps To Reproduce**
- Apply the khcheck manifest for an http-check without specifying the `PASSING_PERCENT` env var.

**Expected behavior**
A default value like 75 or 80 should be automatically set as PASSING_PERCENT if it is not provided as an env var.

**Screenshots**

**Versions**

**Additional context**
It's a very minimal change of adding the following snippet to https://github.com/Comcast/kuberhealthy/blob/master/cmd/http-check/main.go
```go
if len(passing)==0 {
    passing = "100" //or 80 or whatever the default passing percent value
}
```

yashvardhan-kukreja:
closes #780

### DIFF
--- a/cmd/http-check/main.go
+++ b/cmd/http-check/main.go
@@ -67,6 +67,10 @@ func init() {
 	// Check that the EXPECTED_STATUS_CODE environment variable is valid.
 	if len(expectedStatusCode) == 0 {
 		expectedStatusCode = "200"
+	
+	// Check that the PASSING_PERCENT environment variable is valid.
+	if len(passing) == 0 {
+		passing = "100"
 	}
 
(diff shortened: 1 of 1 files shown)


integrii:
I think you missed a curly brace here, but we had someone internally do the same fix and they ran it through tests.  I am closing this one in favor of #786.  Thanks for your help reguardless!

PR #783, opened 2021-01-19
Deployment check tol fix

jdowni000:
Deployment check fix that now has the ability to accept multiple tolerations. Additionally, it now checks for the toleration to be  null and does not create any tolerations. I have tested this in lab and has accepted 3 tolerations, single tolerations, and no tolerations. All tests passed and scheduled as expected.


cmd/deployment-check/Makefile
```diff
@@ -1,5 +1,5 @@
 IMAGE="kuberhealthy/deployment-check"
-TAG="v1.7.1"
+TAG="v1.8.2"
```
jonnydawg, line 2: how did this go from 1.7.1 to 1.8.2?? 
when you build and test images, make sure you put them in your own repo / with `-alpha`
We never released 1.8.0 so go ahead and delete the 1.8.0+ images from docker https://hub.docker.com/repository/docker/kuberhealthy/deployment-check

When you build and push like this, you're releasing test images to the public -- be careful please

cmd/deployment-check/README.md
```diff
@@ -74,7 +74,7 @@ spec:
   podSpec:
     containers:
       - name: deployment
-        image: kuberhealthy/deployment-check:v1.7.1
+        image: kuberhealthy/deployment-check:v1.8.2
```
jonnydawg, line 77: same with the versioning here -- too many jumps

cmd/deployment-check/main.go
```diff
@@ -272,3 +274,15 @@ func reportToKuberhealthy(ok bool, errs []string) {
 	}
 	return
 }
+
+// ReportFailureAndExit logs and reports an error to kuberhealthy and then exits the program.
+// If a error occurs when reporting to kuberhealthy, the program fatals.
+func ReportFailureAndExit(err error) {
```
jonnydawg, line 280: what is the need of this func?

go.sum
```diff
@@ -14,15 +14,20 @@ cloud.google.com/go/pubsub v1.0.1/go.mod h1:R0Gpsv3s54REJCy4fxDixWD93lHJMoZTyQ2k
 cloud.google.com/go/storage v1.0.0/go.mod h1:IhtSnM/ZTZV8YYJWCY8RULGVqBDmpoyjwiyrjsg+URw=
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
 github.com/Azure/go-autorest/autorest v0.9.0/go.mod h1:xyHB1BMZT0cuDHU7I0+g046+BFDTQ8rEZB0s4Yfa6bI=
+github.com/Azure/go-autorest/autorest v0.9.6 h1:5YWtOnckcudzIw8lPPBcWOnmIFWMtHci1ZWAZulMSx0=
```
jonnydawg, line 17: why are there so many Azure packages added?

jonnydawg (changes requested):

cmd/deployment-check/deployment.go
```diff
@@ -258,25 +298,81 @@ func createDeployment(ctx context.Context, deploymentConfig *v1.Deployment) chan
 	return createChan
 }
 
+// multipleToelrations splits multiple toleration values entered from spec file as a string
+func multipleTolerations(splitEnvVars []string) ([]corev1.Toleration, error) {
+
+	if splitEnvVars == nil {
+		err := errors.New("nil was entered into multiple toleration parser")
+		ReportFailureAndExit(err)
+	}
```
joshulyne, line 307: you do a check in the tolerationHandler func that makes sure the list has more than 1 item:
```
if len(splitEnvVars) > 1 {
```
so would this check `if splitEnvVars == nil` be necessary?

cmd/deployment-check/README.md
```diff
@@ -110,20 +110,26 @@ spec:
...
-        image: kuberhealthy/deployment-check:v1.6.2
-        imagePullPolicy: IfNotPresent
+        image: kuberhealthy/deployment-check:v1.8.0
+        imagePullPolicy: Always
         env:
           - name: CHECK_DEPLOYMENT_REPLICAS
             value: "6"
           - name: CHECK_DEPLOYMENT_ROLLING_UPDATE
             value: "true"
+          - name: TOLERATION_VALUE
```
jonnydawg, line 120: You should leave this out if the value is just an empty string

cmd/deployment-check/README.md
```diff
@@ -110,20 +110,26 @@ spec:
...
+            value: ""
+          - name: NODE_SELECTOR
+            value: ""
         resources:
           requests:
-            cpu: 15m
+            cpu: 25m
```
jonnydawg, line 126: any reason cpu was upped here?

cmd/deployment-check/deployment-check.yaml
```diff
@@ -10,15 +10,15 @@ spec:
...
+      image: kuberhealthy/deployment-check:v1.8.0
       imagePullPolicy: IfNotPresent
       env:
         - name: CHECK_DEPLOYMENT_REPLICAS
           value: "4"
         - name: CHECK_DEPLOYMENT_ROLLING_UPDATE
           value: "true"
         - name: TOLERATION_VALUE
-          value: "dedicated"
+          value: "dedicated=kuberhealthy"
```
jonnydawg, line 21: These shouldn't be included in the example yaml because not everyone user will have these values, taints, and selectors

jonnydawg (changes requested):

jonnydawg:
I am working on an edit right now

cmd/deployment-check/Makefile
jdowni000: done

cmd/deployment-check/README.md
jdowni000: done

cmd/deployment-check/main.go
jdowni000: I was fatl and reporting to kuberhealthy in the event of an error in multitolerations...however I can handle the error instead and set a default toleration and remove this func

cmd/deployment-check/deployment-check.yaml
jdowni000: I will remove these for my final push then. However I need these left in for testing purposes as my spec file I apply

go.sum
jdowni000: I do not have answer for this. I do not believe I imported any new packages for this check revision

cmd/deployment-check/deployment.go
jdowni000: I received feedback to always have something like this in to make sure that no one is abusing the function. I understand it is repetitive. Let me know how you want me to proceed.

cmd/deployment-check/README.md
jdowni000: we will be using this env variable. If I remove it completely someone would need to read through the code and add this manually to their spec file to use. My code already looks for this to be empty and assigns a default nil in the event it is. Let me know how you want me to proceed.

cmd/deployment-check/README.md
jdowni000: no. I believe some of these changes we are seeing came from pulling and merging master to keep my code up to date. I will move it back to 15m

cmd/deployment-check/README.md
jonnydawg: this is why we have a list of configurable things in the readme

not everyone is going to be using the environment variable -- we use it yes, but we can modify those values in the spots where we use them

jdowni000: understand. Thank you

cmd/deployment-check/deployment.go
joshulyne: thats fair, i wasn't thinking that it was a function that could be reused! 👍

Issue #794, opened 2021-01-26
Get extraAnnotations from `KuberhealthyCheck` and set it to the pod that is actually for checker


Sn0rt:
1. set `extraAnnotations` for KuberhealthyCheck

I have set the extra annotation with  KuberhealthyCheck resource

```yaml
# kubectl  get khchecks.comcast.github.io daemonset -o yaml
apiVersion: comcast.github.io/v1
kind: KuberhealthyCheck
metadata:
  annotations:
  name: daemonset
  namespace: kuberhealthy
spec:
  extraAnnotations:
    v2-subnet: 7832d617-23ea-42e6-a5a3-efa92435a7c8
    v2-tenant: dcf722f63d0249f8b154876a79e1ce05
    v2-vpc: f203cacb-d15d-473d-9092-48c809f9f618
  podSpec:
    containers:
    - env:
      - name: POD_NAMESPACE
        valueFrom:
          fieldRef:
            fieldPath: metadata.namespace
      - name: CHECK_POD_TIMEOUT
        value: 10m
      image: kuberhealthy/daemonset-check:v3.2.3
      imagePullPolicy: IfNotPresent
      name: main
      resources:
        requests:
          cpu: 10m
          memory: 50Mi
      securityContext:
        allowPrivilegeEscalation: false
        readOnlyRootFilesystem: true
    securityContext:
      fsGroup: 999
      runAsUser: 999
    serviceAccountName: daemonset-khcheck
  runInterval: 15m
  timeout: 12m
```

2. get pod 

get the pod YAML by the command, `# kubectl  get pod daemonset-1611628373 -o yaml`

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    comcast.github.io/check-name: daemonset
    v2-subnet: 7832d617-23ea-42e6-a5a3-efa92435a7c8
    v2-tenant: dcf722f63d0249f8b154876a79e1ce05
    v2-vpc: f203cacb-d15d-473d-9092-48c809f9f618
  creationTimestamp: "2021-01-26T02:32:58Z"
  labels:
    app: kuberhealthy-check
    kuberhealthy-check-name: daemonset
    kuberhealthy-run-id: 77a3f35e-c6bf-4368-9f7d-ec329b66a0e0
  name: daemonset-1611628373
  namespace: kuberhealthy
  ownerReferences:
  - apiVersion: v1
    kind: Pod
    name: kuberhealthy-54cd797c66-bzkh9
[...]

PR #795, opened 2021-01-26
opt: set extraAnnotations to running pod

Sn0rt:
#794 

Signed-off-by: sn0rt <sn0rt@abc.shop.edu.cn>

### DIFF
--- a/cmd/daemonset-check/run_check.go
+++ b/cmd/daemonset-check/run_check.go
@@ -336,6 +336,17 @@ func generateDaemonSetSpec() *appsv1.DaemonSet {
 	if err != nil {
 		log.Errorln("Error getting ownerReference:", err)
 	}
+	if len(ownerRef) == 0 {
+		log.Errorln("Error the length of ownerReference is 0")
+	}
+
+	ownerPod, err := client.CoreV1().Pods(checkNamespace).Get(context.Background(), ownerRef[0].Name, metav1.GetOptions{})
+	if err != nil {
+		log.Errorln("Error getting ownerReference POD", err)
+	}
+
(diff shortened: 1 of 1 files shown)


cmd/daemonset-check/run_check.go
```diff
@@ -337,6 +337,14 @@ func generateDaemonSetSpec() *appsv1.DaemonSet {
 		log.Errorln("Error getting ownerReference:", err)
 	}
 
+	ownerPod, err := client.CoreV1().Pods(checkNamespace).Get(context.Background(), ownerRef[0].Name, metav1.GetOptions{})
```
integrii, line 340: Can you add a check above for `len(ownerRef) > 0` please?

integrii, line 340: Also, because `util.GetOwnerRef` returns the owner of the pod that the code runs on, we can expect `ownerRef` to be the name of the Kuberhealthy pod that spawned this daemonset pod.  Later on when we set the daemonset annotations to `ownerPod.Annotations`, we are actually setting the annotations to whatever the _parent_ Kuberhealthy pod has. 

I think what we want here is to set the annotations and owner reference to whatever the daemonset check pod has.

integrii (changes requested):

cmd/daemonset-check/run_check.go
Sn0rt: sure

integrii:
I am going to close this one for now.  Check on the conversation in #794 and please re-open after.  I think we want this ability, but a slightly different implementation of how its configured.

PR #800, opened 2021-01-27
Update helm-repo-update.sh

integrii:
This changes our helm chart so that when a tag is made for the project, we up the appVersion in the chart.  Also, whenever any change to the chart is merged to master, the automation will build a new unique helm chart version.  This means that the helm chart is no longer published at the same time as kuberhealthy.


jdowni000 (approved):
lgtm

PR #829, opened 2021-02-12
add err to errMessages and update required role

### LINKED ISSUE #828: pod-restarts-check silently fails since role needs permissions to list events
**Describe the bug**
`pod-restarts-check` is silently failing (ie false negatives) and it always reports as healthy if `doChecks` fails. 

**Steps To Reproduce**
- Mis-configure the rbac role
- Run the check
- See that no err is submitted

**Expected behavior**
The check should report failure to the reporting service

**Versions**
 - Cluster OS: amzn-linux
 - Kubernetes Version: v1.18.9-eks-d1db3c
 - Kuberhealthy v2.3.1

isaaguilar:
- fixed updating `errorMessages` with  `err` 
- fixed pod-restarts-check rbac role to be able to list events

Fixes #828 

Signed-off-by: Isa Aguilar <isa@aguilarstory.com>

### DIFF
--- a/cmd/pod-restarts-check/main.go
+++ b/cmd/pod-restarts-check/main.go
@@ -143,6 +143,10 @@ func (prc *Checker) Run() error {
 	case err := <-doneChan:
 		if len(prc.BadPods) != 0 || err != nil {
 			var errorMessages []string
+			if err != nil {
+				log.Error(err)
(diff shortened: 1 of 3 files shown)


integrii (approved):
lgtm!

Thanks for finding and fixing this one!

Issue #861, opened 2021-03-10
Make reaper have configurable reap time on Errored pods


integrii:
The check reaper leaves pods that have errored for five days [statically in the code](https://github.com/Comcast/kuberhealthy/blob/5c8da01d1ef853c63b7a815d896373fbfed21a87/cmd/kuberhealthy/reaper.go#L166-L168).  It would be nice to make this configurable by the user so they aren't forced to have five days of pods in `Error` laying around.

Issue #859, opened 2021-03-04
jobCleanupDuration and maxCheckPods are not properly respected


AshutoshNirkhe:
**Describe the bug**
Currently Kuberhealthy (reaper part) doesn't seem to be properly working. It does reap the pods but seems like lowering the values of  jobCleanupDuration and maxCheckPods configs doesn't really help much.

**Steps To Reproduce**
- Set below values either through values.yaml or just edit the kuberhealthy configmap,
    jobCleanupDuration: 3m
    maxCheckPods: 1
- Optionally restart Kuberhealthy pods to make sure they pick up the changes.

We still see at least 2 pods of each checker whatsoever tweaks I make to those 2 values,

$  kubectl get pods -n synthetic
NAME                             READY   STATUS      RESTARTS   AGE
deployment-1614851948            0/1     Completed   0          94m
deployment-1614855548            0/1     Completed   0          34m
dns-status-internal-1614857048   0/1     Completed   0          9m11s
dns-status-internal-1614857348   0/1     Completed   0          4m11s
kuberhealthy-9d594c6f5-hrf5b     1/1     Running     0          4h35m
kuberhealthy-9d594c6f5-khxxd     1/1     Running     0          4h36m

**Expected behavior**
We would ideally like to get rid off all 0/1 Completed pods to be reaped as much quickly as possible to free up IPs. But at least if we could lower this down to 1 pod per check, it will be really awesome!
PS, I tried with jobCleanupDuration=1m and maxCheckPods=0 , but it didn't help.

**Screenshots**
If applicable, add screenshots to help explain your problem.  You can paste images into github and they will upload automatically.

**Versions**
 - Kubernetes Version: 1.18.14
 - Kuberhealthy Release v2.4.0

**Additional context**

AshutoshNirkhe:
Ok, small correction. It does work at first and then sometimes (may be its something to do with repaer logic frequency of checking) but the point is it keeps the Completed pods hanging around which doesn't really add any value but keeps using IPs,

Settings -
    jobCleanupDuration: 1m
    maxCheckPods: 0

$  kubectl get pods -n synthetic
NAME                             READY   STATUS      RESTARTS   AGE
deployment-1614859236            0/1     Completed   0          12m
deployment-1614859836            0/1     Completed   0          2m15s
dns-status-internal-1614859836   0/1     Completed   0          2m15s
kuberhealthy-69ff96ff5c-gk659    1/1     Running     1          4d21h
[...]

Issue #857, opened 2021-02-25
Cannot unmarshal string into Go struct field Config.JobCleanupDuration of type time.Duration


adriananeci:
**Describe the bug**

Kuberhealthy cannot parse the configmap when `jobCleanupDuration` option is defined because it cannot unmarshal string into Go struct field `Config.JobCleanupDuration` of type `time.Duration`

```
time="2021-02-25T05:12:29Z" level=info msg="WARNING: Failed to read configuration file from disk: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field Config.JobCleanupDuration of type time.Duration"
```

**Steps To Reproduce**
- Start kuberhealthy with the following associated configmap
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: kuberhealthy
  namespace: test
data:
  kuberhealthy.yaml: |-
    listenAddress: ":8080" # The port for kuberhealthy to listen on for web requests
    enableForceMaster: false # Set to true to enable local testing, forced master mode
    logLevel: "debug" # Log level to be used
    listenNamespace: "test" # Kuberhealthy will only monitor khcheck resources from this namespace
    jobCleanupDuration: 3m # The maximum age of khjobs before being reaped
    maxCheckPods: 1 # The maximum number of check pods in Completed state before being reaped
```
- Check the logs and you'll see a warning like
```
time="2021-02-25T05:12:29Z" level=info msg="WARNING: Failed to read configuration file from disk: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field Config.JobCleanupDuration of type time.Duration"
```
- Try to adjust `jobCleanupDuration` value in configmap
- Check kuberhealthy logs and you'll see
```
time="2021-02-25T10:07:35Z" level=debug msg="configReloader: configuration file hash has changed to: 1807f3c6108d085b5b5b6b2a87294b16"
time="2021-02-25T10:07:41Z" level=error msg="configReloader: Error reloading config: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field Config.JobCleanupDuration of type time.Duration"
```

**Expected behavior**
No warnings/errors during starting or configmap changes 

**Versions**
 - Cluster OS: Flatcar
 - Kubernetes Version: 1.18.14
 - Kuberhealthy Release: v2.4.0

integrii:
[...]

Issue #655, opened 2020-09-22
Re-enable unit tests in Kuberhealthy dockerfile


integrii:
Kuberhealthy's container build currently is missing unit tests.  The below line is commented out:

```
#RUN go test -v -short -- --debug --forceMaster
```

We should re-enable this and ensure all tests pass.  Bonus points for fixing existing tests instead of disabling them.

jonnydawg:
I think we should do this as a github action instead -- maybe append a job/task to the gosec action?

jonnydawg:
I've added a gotest action on the `gotest` branch:

Here are the logs from running gotest:
```
[3:02:52](gotest) % go test ./...                                                                                                                                                                                                ~/go/src/github.com/Comcast/kuberhealthy
?       github.com/Comcast/kuberhealthy/v2/cmd/ami-check        [no test files]
?       github.com/Comcast/kuberhealthy/v2/cmd/check-reaper     [no test files]
?       github.com/Comcast/kuberhealthy/v2/cmd/daemonset-check  [no test files]
ok      github.com/Comcast/kuberhealthy/v2/cmd/deployment-check 0.028s
ok      github.com/Comcast/kuberhealthy/v2/cmd/dns-resolution-check     0.148s
?       github.com/Comcast/kuberhealthy/v2/cmd/http-check       [no test files]
?       github.com/Comcast/kuberhealthy/v2/cmd/http-content-check       [no test files]
?       github.com/Comcast/kuberhealthy/v2/cmd/kiam-check       [no test files]
time="2020-09-22T15:03:38-07:00" level=warning msg="Failed to open namespace file: open /var/run/secrets/kubernetes.io/serviceaccount/namespace: no such file or directory"
time="2020-09-22T15:03:38-07:00" level=info msg="Did not find instance namespace. Using default namespace: kuberhealthy"
time="2020-09-22T15:03:38-07:00" level=info msg="Kuberhealthy is located in the kuberhealthy namespace."
time="2020-09-22T15:03:38-07:00" level=fatal msg="Failed to build kubernetes configuration from configuration flags"
FAIL    github.com/Comcast/kuberhealthy/v2/cmd/kuberhealthy     0.031s
?       github.com/Comcast/kuberhealthy/v2/cmd/network-connection-check [no test files]
ok      github.com/Comcast/kuberhealthy/v2/cmd/pod-restarts-check       0.028s [no tests to run]
[...]

PR #862, opened 2021-03-10
Fix checkReaper duration configs, add failedPodCleanupDuration, re-enable tests

joshulyne:
Addresses #857 #861
and possibly #859, #655

### DIFF
--- a/cmd/kuberhealthy/reaper.go
+++ b/cmd/kuberhealthy/reaper.go
@@ -34,9 +34,7 @@ const maxCheckerPodsDefault = 4
 const jobCleanupDurationDefault = time.Minute * 15
 const failedPodCleanupDurationDefault = time.Hour * 120
 
-// Parse duration configs into time.Duration. The way we supply the duration configs in the Kuberhealthy config, ie 15m, is interpreted
-// as a string instead of time.Duration when we unmarshal the json in the config reloader. This fails since since Go marshals time.Duration
-// as an int64 and not string.
+
 var jobCleanupDuration = jobCleanupDurationDefault
 var failedPodCleanupDuration = failedPodCleanupDurationDefault
 
@@ -49,6 +47,9 @@ func init() {
 
 }
 
+// parseDuration parses checkReaper duration configs into time.Duration. Go parses time.Duration from int64 and not from
+// string, which is how the config reloader unmarshals the config json. This way we can still supply time.Duration type
+// values in the config, instead of an int64 that doesn't allow users to specify the duration type.
 func parseDuration() {
 
 	var err error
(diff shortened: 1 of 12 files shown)


cmd/kuberhealthy/reaper.go
```diff
@@ -40,8 +44,26 @@ func init() {
...
+// parseDuration parses checkReaper duration configs into time.Duration. Go parses time.Duration from int64 and not from
+// string, which is how the config reloader unmarshals the config json. This way we can still supply time.Duration type
+// values in the config, instead of an int64 that doesn't allow users to specify the duration type.
+func parseDuration() {
+
+	var err error
+	if len(cfg.JobCleanupDuration) != 0 {
+		jobCleanupDuration, err = time.ParseDuration(cfg.JobCleanupDuration)
+		if err != nil {
+			log.Errorln("error occurred attempting to parse JobCleanupDuration:", err, "Using default value:", jobCleanupDuration)
```
integrii, line 58: In both cases where we run `time.ParseDuration` here, we should be certain what is returned when an error occurs.  If `time.ParseDuration` returns a `0`, this could cause some real chaos.  Consider setting the default explicitly after an error occurs.

cmd/kuberhealthy/reaper.go
```diff
@@ -40,8 +44,26 @@ func init() {
...
 	}
 
-	if cfg.JobCleanupDuration == 0 {
-		cfg.JobCleanupDuration = jobCleanupDurationDefault
+}
+
+// parseDuration parses checkReaper duration configs into time.Duration. Go parses time.Duration from int64 and not from
+// string, which is how the config reloader unmarshals the config json. This way we can still supply time.Duration type
+// values in the config, instead of an int64 that doesn't allow users to specify the duration type.
+func parseDuration() {
```
integrii, line 52: Consider making this a method on the `Config` struct.  This currently sits in the global namespace for the binary.

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -149,7 +151,7 @@ check:
     image:
       registry: kuberhealthy
       repository: dns-resolution-check
-      tag: v1.4.2
+      tag: v1.4.3
```
integrii, line 154: be sure this image exists before we merge this to master

integrii (changes requested):

deploy/helm/kuberhealthy/values.yaml
joshulyne: this was updated for the external-dns-check but not the internal-dns-check :/
https://github.com/Comcast/kuberhealthy/pull/862/files/9e4b4297476f925fcdec45e9b79b390d7e804f1c#diff-a1d3641891145c247db5c0446a75d77fc036d916b0d26d6efe258424f3e7f440R174

integrii (approved):

PR #869, opened 2021-03-18
fix: Set proper pod port where prometheus metrics are exposed

randrusiak:
Currently, prometheus-server tries to scrape metrics from pod's port 80 which isn't exposed.
According to default prometheus [kubernetes_sd_config](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L1592) service annotation `prometheus.io/port` should point to pod's port instead of service port.

### DIFF
--- a/deploy/helm/kuberhealthy/templates/service.yaml
+++ b/deploy/helm/kuberhealthy/templates/service.yaml
@@ -10,7 +10,7 @@ metadata:
   annotations:
   {{- if and .Values.prometheus.enabled (not .Values.prometheus.serviceMonitor.enabled) }}
     prometheus.io/scrape: "true"
-    prometheus.io/port: "80"
+    prometheus.io/port: "8080"
     prometheus.io/path: "/metrics"
   {{- end }}
   {{- range $key, $value := .Values.service.annotations }}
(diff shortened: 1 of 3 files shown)


deploy/kuberhealthy-prometheus.yaml
```diff
@@ -321,7 +321,7 @@ metadata:
   namespace: kuberhealthy
   annotations:
     prometheus.io/scrape: "true"
-    prometheus.io/port: "80"
+    prometheus.io/port: "8080"
```
integrii, line 324: this will get regenerated by our build automation when the helm chart is updated - fyi

deploy/helm/kuberhealthy/templates/service.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
   annotations:
   {{- if and .Values.prometheus.enabled (not .Values.prometheus.serviceMonitor.enabled) }}
     prometheus.io/scrape: "true"
-    prometheus.io/port: "80"
+    prometheus.io/port: "8080"
```
integrii, line 13: @joshulyne can you verify 8080 is the right port here??

joshulyne: yupp, he's right!

joshulyne (approved):

PR #871, opened 2021-03-25
Add khcheck-external-secrets to registry document

Nick-Triller:
Hi,

you can see the [rendered markdown file here](https://github.com/Nick-Triller/kuberhealthy/blob/document-khcheck-external-secrets/docs/EXTERNAL_CHECKS_REGISTRY.md). I also auto-formatted the file with VS Code before adding the new entry.

Please note I linked to a helm chart instead of rendered manifests. The reason for this is the ExternalSecret that is created by the check will have to be customized in most cases.


integrii (approved):
Hey Nick.  This is awesome!  Thanks for adding your check here.  Your check is really well documented and your contribution is really appreciated.  I am thrilled to see more checks entering the world.

Issue #858, opened 2021-02-26
Possible race condition while updating khstate


adriananeci:
**Describe the bug**
Possible race condition while updating khstate.
Most probably a check failed few days ago due to `Reporting took too long to complete and timed out.` and was never scheduled anymore since then.
From kuberhealthy logs:
```
[kuberhealthy-9f49c5bf9-dg5vl kuberhealthy] time="2021-02-26T12:13:58Z" level=info msg="ee01135b-6584-44a0-a996-9cb131eb80d8 e2e-test/e2e-k8sapi: [Failed to write new UUID for check because object was modified by another process.  Retrying in 5s]"
[kuberhealthy-9f49c5bf9-dg5vl kuberhealthy] time="2021-02-26T12:14:03Z" level=info msg="ee01135b-6584-44a0-a996-9cb131eb80d8 e2e-test/e2e-k8sapi: [Failed to write new UUID for check because object was modified by another process.  Retrying in 5s]"
[kuberhealthy-9f49c5bf9-dg5vl kuberhealthy] time="2021-02-26T12:14:08Z" level=info msg="ee01135b-6584-44a0-a996-9cb131eb80d8 e2e-test/e2e-k8sapi: [Failed to write new UUID for check because object was modified by another process.  Retrying in 5s]"
[...]

PR #872, opened 2021-03-29
fix for retries in #858

integrii:
This should fix the retry infinite loop when a race condition happens while updating the UUID of a checker pod on the `khcheck` resource.


integrii:
This is in testing in our test cluster, but is looking good and expected to merge soon.

jonnydawg (approved):
thanks man

joshulyne (approved):
lgtm 👍

PR #877, opened 2021-03-31
Checks registry update

integrii:
- add SSL handshake check to registry
- add SSL expiration check to registry
- change incorrect check application instructions at the top of the registry page


jdowni000 (approved):
lgtm

Issue #870, opened 2021-03-22
Kuberhealthy Bad Request on GKE (In-cluster inter-pod SNAT behaviour)


DannyLyubenov:
**Describe the bug**
`level=fatal msg="error when reporting to kuberhealthy: bad status code from kuberhealthy status reporting url: [400] 400 Bad Request`.
- Kuberhealthy responding with Bad Request after a successful HTTP check.
- This might be down to GKE using NAT for internal routing.
- Related issue - https://github.com/Comcast/kuberhealthy/issues/866
**Steps To Reproduce**
- Obtain Kuberhealthy yaml manifest:
```wget https://raw.githubusercontent.com/Comcast/kuberhealthy/master/deploy/kuberhealthy.yaml```
- Remove default Kuberhealthy checks from the manifest above (truncated for truncated):
```
---
# Source: kuberhealthy/templates/khcheck-daemonset.yaml
apiVersion: comcast.github.io/v1
kind: KuberhealthyCheck
metadata:
  name: daemonset
  namespace: kuberhealthy
spec:
  ...
---
# Source: kuberhealthy/templates/khcheck-deployment.yaml
apiVersion: comcast.github.io/v1
kind: KuberhealthyCheck
metadata:
  name: deployment
  namespace: kuberhealthy
spec:
  ...
---
# Source: kuberhealthy/templates/khcheck-dns-internal.yaml
apiVersion: comcast.github.io/v1
kind: KuberhealthyCheck
metadata:
  name: dns-status-internal
  namespace: kuberhealthy
spec:
  ...
```
- Add to the aforementioned manifest a permissive `PodSecurityPolicy` and associate it to a `ClusterRole` (a GKE requirement, output truncated):
```
---
# Source: kuberhealthy/templates/clusterrole.yaml
apiVersion: "rbac.authorization.k8s.io/v1"
kind: ClusterRole
metadata:
  name: kuberhealthy
rules:
  ...
  - apiGroups: ['policy']
    resources: ['podsecuritypolicies']
    verbs:     ['use']
    resourceNames:
    - kuberhealthy-psp
---
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: kuberhealthy-psp
spec:
  privileged: true
  allowPrivilegeEscalation: true
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  runAsUser:
    rule: RunAsAny
  fsGroup:
[...]

PR #884, opened 2021-04-13
Refactor external report validation to use check run uuid, allow 0 maxCompletedCheckPods in checkReaper

joshulyne:
Addresses #859 
CheckReaper:
1) checkReaper: Rename maxCheckPods --> maxCompletedCheckPods (maximum number of completed pods in Completed state before getting reaped, excluding current check run pod)
2) Allow `maxCompletedCheckPods = 0` --> implies that only current completed check run pod remains

Addresses #870
External Request Validation:
External request is validated first by looking for the request header value: `kh-run-uuid`. If header is not set, or validation fails, validate using pod remote IP (old method)
Added Logic:
- get env var `KH_RUN_UUID` and set it as a header `kh-run-uuid` in the send report request
- in the external report validation, get the UUID or value from this header and use it to fetch the pod using the label: `kuberhealthy-run-id` + UUID. 
- verify pod has env vars, and check for the `KH_RUN_UUID` env var
- validate the env var UUID with khstate current UUID 👍


cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1171,15 +1164,15 @@ func (k *Kuberhealthy) validateExternalRequest(remoteIPPort string) (PodReportIP
...
+	log.Debugln("Env vars found on pod with kh run uuid", uuid, envVars)
 
 	// validate that the environment variables we expect are in place based on the check's name and UUID
 	// compared to what is in the khcheck custom resource
 	var foundUUID bool
 	for _, e := range envVars {
 		log.Debugln("Checking environment variable on calling pod:", e.Name, e.Value)
 		if e.Name == external.KHRunUUID {
-			log.Debugln("Found value on calling pod", ip, "value:", external.KHRunUUID, e.Value)
+			log.Debugln("Found value on calling pod", uuid, "value:", external.KHRunUUID, e.Value)
```
joshulyne, line 1175: kept most of this logic in here -- sort of seems repetitive since the uuid set in the header is from the environment variable and we validate it again here, thoughts? but when you go further down, this uuid grabbed from the env var here is compared to its khstate current run UUID which I think is important @integrii

integrii:
Because we will have some period of time where checks haven't been recompiled, but kuberhealthy will be running this new code, we should think about what we do when the UUID header is missing. 

It seems like we should fall back to the old way of validation through the kubernetes API when the header is missing to avoid a complex upgrade.  Then, later on in a future version, we fully remove the API validation.

Thanks for writing this!  I am going to try and look at the code from my phone while I wait at Jury Duty.

joshulyne:
> Because we will have some period of time where checks haven't been recompiled, but kuberhealthy will be running this new code, we should think about what we do when the UUID header is missing.
> 
> It seems like we should fall back to the old way of validation through the kubernetes API when the header is missing to avoid a complex upgrade. Then, later on in a future version, we fully remove the API validation.
> 
> Thanks for writing this! I am going to try and look at the code from my phone while I wait at Jury Duty.

Yeah that was the part I was wondering how to handle -- https://github.com/Comcast/kuberhealthy/pull/884/files#diff-6fe2eb1bda3003e83c281ea43553e293915389e31314d523116afe24efe877e3R1291-R1298
^^ that line grabs that header value and checks if it exists... right now, if the label isn't set we just return and continue and the check just fails with your `check timed out waiting for report` error -- which i'm not sure is the best way to handle that. I think we have to be careful with how we want kuberhealthy to fallback on this other way of validating external reports because it seems like for some of our users that's not even an option, but for others and for compatibility purposes, that is needed. I do know we have to rebuild all of our checks to get the updated checkClient package.

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1292,20 +1291,46 @@ func (k *Kuberhealthy) externalCheckReportHandler(w http.ResponseWriter, r *http
...
-	if err != nil {
-		w.WriteHeader(http.StatusBadRequest)
-		k.externalCheckReportHandlerLog(requestID, "Failed to look up pod by IP:", r.RemoteAddr, err)
-		return nil
+	// Get kh-run-uuid from the request header and verify it exists. If the kh-run-uuid doesn't exist, verify request by
+	// validating using the pod's remote IP.
+	reportingKHRunUUID := r.Header.Get("kh-run-uuid")
+	var podReport PodReportInfo
+	var err error
+	if len(reportingKHRunUUID) == 0 {
```
integrii, line 1301: function all this logic out into something like this:

```go
var reportValidated bool
var podReport PodReportInfo
var err error
if requestUsesUUIDHeader(r) {
      podReport, err = ValidateCheckUUID(r.Header.Get("kh-run-uuid"))
      if err != nil {
          // log the error we had validating
      } else {
          // if no issue validating, set a bool so that we don't try the legacy IP resolution checks
          checkValidated = true
      }
}
```

Then, drop the else below this code for something like this:

```go
if !reportValidated {
  // try to validate pod by searching for its source it (legacy behavior)
  podReport, err = ValidateCheckUUIDBySourceIP(r)
  if err != nil {
      // throw error back to client and log it. both header and source ip validation have failed
  }
}
```

go.mod
```diff
@@ -19,9 +20,9 @@ require (
 	github.com/sirupsen/logrus v1.4.0
 	github.com/smartystreets/goconvey v1.6.4 // indirect
 	gopkg.in/ini.v1 v1.51.0 // indirect
-	gopkg.in/yaml.v2 v2.2.8
+	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.19.3
-	k8s.io/apimachinery v0.19.3
+	k8s.io/apimachinery v0.21.0
```
integrii, line 25: did you mean to bump `apimachinery` up to 21 with `api` and `client-go` still at 19?

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -1292,20 +1291,46 @@ func (k *Kuberhealthy) externalCheckReportHandler(w http.ResponseWriter, r *http
...
+		k.externalCheckReportHandlerLog(requestID, "Received blank kh-run-uuid header from external request. Validating external check status report from the pod's remote IP.")
+
+		// break the port off the RemoteAddr incoming string
+		ipPortString := strings.Split(r.RemoteAddr, ":")
+		if len(ipPortString) == 0 {
```
integrii, line 1306: Whenever I see a `len()` check on a slice, I wonder if I should instead be fetching the slice from some function that is smart enough to return an error.  This way errors are explicit values, rather than ascertained by the length of the slice in your calling code.

cmd/kuberhealthy/kuberhealthy.go
integrii: Just remember that we want to try to validate with the header first because it does not require the extra call to the Kube API.  We fall back to doing a query for pods by IP _only_ if the header validation fails.

integrii (changes requested):

cmd/kuberhealthy/reaper.go
```diff
@@ -89,11 +83,11 @@ func reaper(ctx context.Context) {
 
 	log.Infoln("checkReaper: starting up...")
 	log.Infoln("checkReaper: job cleanup duration:", cfg.JobCleanupDuration)
-	log.Infoln("checkReaper: max checker pods:", cfg.MaxCheckPods)
+	log.Infoln("checkReaper: max completed checker pods:", cfg.MaxCompletedCheckPods)
 	log.Infoln("checkReaper: failed pod cleanup duration:", cfg.FailedPodCleanupDuration)
 
 	// start a new ticker
-	t := time.NewTicker(time.Minute * 3)
+	t := time.NewTicker(time.Second * 30)
```
integrii, line 90: Let's make this configurable from an environment variable and default it to `30s` if not configured.  How about the env var `CHECK_REAPER_RUN_INTERVAL` with type of `time.Duration`?

pkg/metrics/exporter.go
```diff
@@ -60,8 +60,12 @@ func GenerateMetrics(state health.State) string {
 		metricCheckState[metricName] = checkStatus
 		runDuration, err := time.ParseDuration(d.RunDuration)
 		if err != nil {
-			log.Errorln("Error parsing run duration:", d.RunDuration, "for metric:", metricName, "error:", err)
-			continue
+			if d.RunDuration != "" {
```
integrii, line 63: do we need this `if` here?  It seems like if there was an error parsing the duration, the `time.ParseDuration` call on 61 would have found it.  At this point the only possible "bad" duration that `runDuration` could be set to would be `0`.  you could just check that, but check it after this `err != nil` if statement.

integrii: Same with the code down on line 90

integrii (changes requested):

cmd/kuberhealthy/reaper.go
joshulyne: feel like it would have to go in the kuberhealthy configmap no? that's where the other checkReaper configurations are being set.

joshulyne: actually, you're right, setting it as an env var is better.

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -3,7 +3,7 @@
 # Declare variables to be passed into your templates.
 
 checkReaper:
-  maxCheckPods: 4 # the maximum number of check pods in Completed state before being reaped
+  maxCompletedCheckPods: 4 # Maximum number of check pods in Completed state before being reaped, excludes pod from the most recent check run. If not set or set to 0, only the most recent check pod will remain.
```
integrii, line 6: We might want to allow people to have no history at all.  In the case that the user configures a `0` here, we probably want to delete anything that is in the state `Completed` or `Error` immediately.

'most recent' in the comment leads me to believe that we always keep one `Completed` pod around, even if one is running.  I don't think that is required.  We could let users have no history at all if they wish.

cmd/kuberhealthy/reaper.go
```diff
@@ -224,7 +227,7 @@ func (k *KubernetesAPI) deleteFilteredCheckerPods(ctx context.Context, client *k
 		// Delete if there are more than MaxPodsThreshold checker pods with the same name in status Succeeded that were created more recently
 		// Delete if the checker pod is Failed and there are more than MaxPodsThreshold Failed checker pods of the same type which were created more recently
 		allCheckPods := getAllPodsWithCheckName(reapCheckerPods, v)
-		if len(allCheckPods) > cfg.MaxCheckPods {
+		if len(allCheckPods) > cfg.MaxCompletedCheckPods {
```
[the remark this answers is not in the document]
integrii: Thinking about it more... this is probably all made more complicated by configuring both a max age for pods and a max count... maybe we should pick one.  Either we delete based on how many pods there are, or we delete after a specific age.  A specific age seems like the easiest way to do it - you wouldn't have to select and delete the oldest pods in that logic... Only the pods that are older than a timestamp.

integrii (changes requested):

deploy/helm/kuberhealthy/values.yaml
AshutoshNirkhe: Sorry for getting involved here like this. But I do agree with @integrii , it will be best if we leave no pods around if we set maxCompletedCheckPods. But may be we still leave 1 error pod around for 'failedPodCleanupDuration' time for debugging the issue in case of Errors.

joshulyne: @AshutoshNirkhe, not at all and thanks for your input! makes sense 👍  will update to adjust for this

cmd/kuberhealthy/reaper.go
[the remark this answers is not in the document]
joshulyne: Not quite, `failedPodCleanupDurationDefault` or the max age for pods apply only to failed pods (pods in error state), not completed pods which is why we have them configured separately. Also, I do think using max pod count is important because if we only specify using a timestamp, checks that run every minute will be will be much more prolific compared to checks that run every hour or later. 

BUT, I do agree these configurations are complicated and there is definitely room for simplification...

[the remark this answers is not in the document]
joshulyne: Currently the logic is 
Failed Pods:
- Keep around for certain duration or UP UNTIL count of failed pods > `maxCompletedCheckPods`

Completed Pods:
- Keep around until count of completed pods > `maxCompletedCheckPods`

I think the reason we want to keep the pod count around is because of the different check run intervals

integrii:
Leaving this for today.  Looks like the new yaml package (`	yaml "gopkg.in/yaml.v2"`) works!

integrii:
This is fixed to not require an extra stage of parsing now! @joshulyne

integrii (approved):
- included renaming some packages labeled `v12` as their better names such as `metav1`.
- refactored a test a little bit because the new yaml parser is faster
- removed parsing process when unmarshalling yaml

AshutoshNirkhe:
@integrii @joshulyne just a quick check here. Is it safe if we make use of latest app version from 'master' ? 
E.g. '63' as of now while referencing kuberhealthy v2.4.1 docker image ? In fact I tested that for getting my podAntiAffinity stuff included in our internal release, which worked fine! But I guess it could be dangerous because you guys haven't bumped the app version further to 64 after merging this :) 
Just wanted to know if its better to wait for official v2.4.2 release instead for both things i.e. podAntiAffinity and maxCompletedCheckPods updates ?

AshutoshNirkhe:
> @integrii @joshulyne just a quick question here. Is it safe if we make use of latest app version from 'master' ?
> E.g. '63' as of now while referencing kuberhealthy v2.4.1 docker image ? In fact I tested that for getting my podAntiAffinity stuff included in our internal release, which worked fine! But I guess it could be dangerous because you guys haven't bumped the app version further to 64 after merging this PR :)
> Just wanted to know if its safer to wait for official v2.4.2 release instead for both things i.e. podAntiAffinity and maxCompletedCheckPods updates ?

I checked now in https://comcast.github.io/kuberhealthy/helm-repos/archives/kuberhealthy-63.tgz and see only podAntiAffinity stuff and not maxCompletedCheckPods stuff, so am assuming probably that's yet to be built. But still I am interested in knowing in general if this practice of using latest app version from master is safe or no :)

integrii:
Master is often not very safe. Being in master means it's passed code review and tests. However, things in master often still need release candidate style testing before they become a release.

We are currently working on rebuilding all checks with new versions that we're built off of the new checkClient package. After that we will begin testing a release that removes the source IP validation logic that is required now (it's already patched out in master).

AshutoshNirkhe:
> Master is often not very safe. Being in master means it's passed code review and tests. However, things in master often still need release candidate style testing before they become a release.
> 
> We are currently working on rebuilding all checks with new versions that we're built off of the new checkClient package. After that we will begin testing a release that removes the source IP validation logic that is required now (it's already patched out in master).

Thanks, I merged the podAntiAffinity (that uses 63 version chart) this morning. But will keep this in mind for our next upgrades.

AshutoshNirkhe:
> > Master is often not very safe. Being in master means it's passed code review and tests. However, things in master often still need release candidate style testing before they become a release.
> > We are currently working on rebuilding all checks with new versions that we're built off of the new checkClient package. After that we will begin testing a release that removes the source IP validation logic that is required now (it's already patched out in master).
> 
> Thanks, I merged the our code that uses podAntiAffinity changes (that uses 63 version chart) this morning. But will keep this in mind for our next upgrades.

AshutoshNirkhe:
> > Master is often not very safe. Being in master means it's passed code review and tests. However, things in master often still need release candidate style testing before they become a release.
> > We are currently working on rebuilding all checks with new versions that we're built off of the new checkClient package. After that we will begin testing a release that removes the source IP validation logic that is required now (it's already patched out in master).
> 
> Thanks, I merged the podAntiAffinity (that uses 63 version chart) this morning. I see you have the app version updated to 64 now, so hopefully we should be good for now. But I will keep this in mind for our next upgrades.

PR #681, opened 2020-10-13
feat: allow users that install pod-status check via helm to optionally allow cluster wide scope

rawlingsj:
related discussion https://github.com/Comcast/kuberhealthy/issues/680

This change also adds a unit test using kubernetes fake client so I needed to change the findPodsNotRunning funtion args, happy to put back and remove the test if prefferred though

PR #897, opened 2021-04-22
feat: optionally allow users to install pod restarts check with a clu…

rawlingsj:
…ster scope

similar to this PR https://github.com/Comcast/kuberhealthy/pull/681

### DIFF
--- a/cmd/pod-restarts-check/pod-restarts-check-clusterscope.yaml
+++ b/cmd/pod-restarts-check/pod-restarts-check-clusterscope.yaml
@@ -1,0 +1,75 @@
+---
+# Source: kuberhealthy/templates/khcheck-pod-restarts.yaml
+apiVersion: comcast.github.io/v1
+kind: KuberhealthyCheck
+metadata:
+  name: pod-restarts
+  namespace: kuberhealthy
+spec:
+  runInterval: 5m
+  timeout: 10m
+  podSpec:
+    securityContext:
+      runAsUser: 999
+      fsGroup: 999
+    containers:
+      - env:
+          - name: MAX_FAILURES_ALLOWED
+            value: "10"
+        image: kuberhealthy/pod-restarts-check:v2.3.3
+        imagePullPolicy: IfNotPresent
+        name: main
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+    restartPolicy: Never
+    serviceAccountName: pod-restart-sa
+---
+# Source: kuberhealthy/templates/khcheck-pod-restarts.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pod-restart-sa
+  namespace: kuberhealthy
+---
+# Source: kuberhealthy/templates/khcheck-pod-restarts.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pod-restart-role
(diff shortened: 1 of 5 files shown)


cmd/pod-restarts-check/pod-restarts-check-clusterscope.yaml
```diff
@@ -0,0 +1,75 @@
...
+  timeout: 10m
+  podSpec:
+    securityContext:
+      runAsUser: 999
+      fsGroup: 999
+    containers:
+      - env:
+          - name: MAX_FAILURES_ALLOWED
+            value: "10"
+        image: kuberhealthy/pod-restarts-check:v2.3.3
```
rawlingsj, line 19: I'm not sure if I need to do anything with this tag?  As this needs to points to a version that hasn't been released yet so maybe the release pipeline updates this?

deploy/helm/kuberhealthy/values.yaml
```diff
@@ -192,6 +192,7 @@ check:
       registry: kuberhealthy
       repository: pod-restarts-check
       tag: v2.3.3
```
rawlingsj, line 194: Same as above, I'm not sure if I need to do anything with this tag?  As this needs to points to a version that hasn't been released yet so maybe the release pipeline updates this?

joshulyne: I believe this version `v2.3.3` has been released? https://hub.docker.com/repository/registry-1.docker.io/kuberhealthy/pod-restarts-check/tags?page=1&ordering=last_updated

hmm we don't have the best release process for specific versions for these external checks -- when a pr for external checks get merged in we do have a github action in place for build and pushing "latest". For specific versions, we build and push manually :/

joshulyne:
@rawlingsj thanks so much for contributing this! will check it out and test it soon!

deploy/helm/kuberhealthy/values.yaml
rawlingsj: ok cool - makes sense, so is it best to change that to latest in this PR and let the followup PR set the correct tag?

cmd/pod-restarts-check/pod-restarts-check-clusterscope.yaml
integrii: This one is actually manual... This is basically documentation.  The versions in the helm chart (and the flat files in the directory with the helm chart) get updated after we change the `version.yaml` of the chart, but this one does not.

deploy/helm/kuberhealthy/values.yaml
integrii: Yep... this is the flow I would expect:

- third party check makes a new release version
- third party check opens a PR to edit this values.yaml file for the helm chart
- helm chart static files are regenerated automatically

joshulyne (approved):
lgtm!

deploy/helm/kuberhealthy/templates/khcheck-pod-restarts.yaml
```diff
@@ -59,6 +61,44 @@ spec:
...
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - list
+{{- end }}
+{{- if not .Values.check.podRestarts.allNamespaces }}
```
joshulyne, line 100: wondering if this could simplified to just an `else`
```suggestion
{{ else }}
```

joshulyne:
@rawlingsj upped the pod-restarts-check to `v2.4.0` which should have this clusterScope feature!

PR #902, opened 2021-04-27
chore: include the namespace in error message for when the pod status…

rawlingsj:
… check is watching all namespaces


joshulyne (approved):
lgtm 👍

PR #935, opened 2021-05-12
Small fixes

joshulyne:
- Fix deployment check yaml restartPolicy indentation
- Remove determineKHWorkload debug log lines
- Retry on checkClient sendReport when status response code isnt 200 or 400
- Return doneChan in shutdown


integrii (approved):
LGTM

Issue #778, opened 2021-01-18
current CRDs aren't valid with the new GA-version of apiextensions.k8s.io/v1


bavarianbidi:
**Describe the bug**
Since kubernetes 1.16 the `apiextensions.k8s.io/v1beta1` is deprecated and will be unavailable in 1.22.
Deploying a CRD with `apiextension.k8s.io/v1beta1` via `kubectl` leads to following warning:

```
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
customresourcedefinition.apiextensions.k8s.io/khjobs.comcast.github.io created
```

**Steps To Reproduce**
- Create a cluster with version `1.19` or `1.20` (`kind create cluster`)
  and run 
  ```
  cat << EOF | kubectl apply -f -
  apiVersion: apiextensions.k8s.io/v1beta1
  kind: CustomResourceDefinition
  metadata:
    name: khstates.comcast.github.io
  spec:
    group: comcast.github.io
    version: v1
    scope: Namespaced
    names:
      plural: khstates
      singular: khstate
      kind: KuberhealthyState
      shortNames:
        - khs
   EOF
  ```

**Expected behavior**
No warning message

**Additional context**

But with some autogenerated CRDs we could use [controller-gen](github.com/kubernetes-sigs/controller-tools/cmd/controller-gen) from Kubebuilder.
Maybe useful to solve https://github.com/Comcast/kuberhealthy/issues/603

following CRD-definition works for now:

```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: khchecks.comcast.github.io
spec:
  group: comcast.github.io
  versions:
  - name: v1
    served: true
    storage: true
    schema:
      openAPIV3Schema:
        type: object
        x-kubernetes-embedded-resource: true
        x-kubernetes-preserve-unknown-fields: true
  scope: Namespaced
  names:
    plural: khchecks
    singular: khcheck
    kind: KuberhealthyCheck
    shortNames:
      - khc
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: khstates.comcast.github.io
spec:
  group: comcast.github.io
  versions:
  - name: v1
    served: true
    storage: true
    schema:
      openAPIV3Schema:
        type: object
[...]

Issue #351, opened 2020-03-12
Include node that checks ran on in `khstate` data


integrii:
It would be really nice if the node that checks ran on was placed into the status JSON.  This way we could programmatically find out which node ran the checker pod easily.

Currently, the only way to tell where a check ran is to check the node assignment on the checker pod with `kubectl`.

integrii:
Bumping to 2.5.0

PR #937, opened 2021-05-12
Generate and update kuberhealthy crds

joshulyne:
Addresses #778 & #351


cmd/kuberhealthy/Makefile
```diff
@@ -1,5 +1,5 @@
 IMAGE="kuberhealthy/kuberhealthy"
-TAG="unstable"
+TAG="crds-update"
```
jonnydawg, line 2: maybe change it back to unstable

pkg/apis/khjob/v1/doc.go
```diff
@@ -1,5 +1,5 @@
 // +k8s:deepcopy-gen=package
 // +k8s:defaulter-gen=TypeMeta
-// +groupName=khjob
+// +groupName=comcast.github.io
```
jonnydawg, line 3: did we still want our apigroup to be `comcast.github.io` or do we want to change it to `kuberhealthy.io`

integrii: this is a breaking change so we'll have to save it for v3...

pkg/checks/external/util/main.go
```diff
@@ -110,7 +110,7 @@ func PodNameExists(client *kubernetes.Clientset, podName string, namespace strin
 	}
 
 	if p.Name == "" {
-		log.Warningln("Pod name is empty. Pod does not exist")
+		log.Warnln("Pod name is empty. Pod does not exist")
```
integrii, line 113: potentially drop the warning here and the one above on line 108?

pkg/health/state.go
```diff
@@ -16,15 +16,17 @@ import (
 	"net/http"
 
 	log "github.com/sirupsen/logrus"
+
+	khstatev1 "github.com/Comcast/kuberhealthy/v2/pkg/apis/khstate/v1"
```
integrii, line 20: Remove `Comcast` here for `kuberhealthy`

integrii (changes requested):

joshulyne:
use camelcase for khstate struct.

integrii (approved):
This is a big one!  Thanks! 🚀

PR #940, opened 2021-05-13
Refactoring all the source links of comcast.github.io -> kuberhealthy.github.io

blame19:
Addresses https://github.com/kuberhealthy/kuberhealthy/issues/938

This replaced the source links  comcast.github.io to kuberhealthy.github.io, plus some Readme changes.


jonnydawg:
Thanks for making this change! Will check it out and report back

integrii:
Thanks for the PR!  The file `helm-repos/index.yaml` reverts some URLs that were fixed, so if you can drop that one from your PR, this would be a great merge.  Some things you found were not changed in the PR I did earlier today.

integrii:
Actually, we probably have this one handled now if you would prefer to just close this one instead of dealing with conflicts and all.

PR #942, opened 2021-05-13
Only restartChecks if pod is master for configReloader

### LINKED ISSUE #925: Checks are getting scheduled at nearly the same time.
**Describe the bug**
A KHC will get scheduled twice -- one after the other within milliseconds:
image
image
image

This always forces the one that is scheduled first to fail / error due to an invalid check uuid resulting in a rejection from the Kuberhealthy master and a `400 Bad Request` from the checker pod.

**Steps To Reproduce**
- Run a deployment of kuberhealthy with `latest` tag
- That container's digest for reference: `sha256:e03e18490bd87e162d2c3a51a7f7f8c6bdd1a883af90b0a7e9a6ea4ef40d1c42` (Today is May 10, 2021)

**Expected behavior**
Normal Kuberhealthy check lifecycle (single checker pod -> check -> report) 

**Versions**
 - Kubernetes Version: 1.11.7

joshulyne:
Fixes #925

### DIFF
--- a/cmd/kuberhealthy/config.go
+++ b/cmd/kuberhealthy/config.go
@@ -167,7 +167,7 @@ func configReloadNotifier(ctx context.Context, notifyChan chan struct{}) {
 
 	// when outChan gets events, reload configuration and checks
 	for range outChan {
-		// if the context has expired, then shut down the master status watcher entirely
+		// if the context has expired, then shut down the config reload notifier entirely
 		select {
 		case <-ctx.Done():
 			log.Debugln("configReloader: stopped notifying config reloads due to context cancellation")
--- a/cmd/kuberhealthy/config.go
+++ b/cmd/kuberhealthy/config.go
@@ -155,7 +155,7 @@ func startConfigReloadMonitoringWithSmoothing(filePath string, scrapeInterval ti
 }
 
 // configReloadNotifier watchers for events in file, reloads the configuration, and notifies upstream to restart checks
-func configReloadNotifier(notifyChan chan struct{}) {
+func configReloadNotifier(ctx context.Context, notifyChan chan struct{}) {
 
 	outChan, cancelFunc, err := startConfigReloadMonitoring(configPath)
 	if err != nil {
@@ -167,6 +167,14 @@ func configReloadNotifier(notifyChan chan struct{}) {
 
 	// when outChan gets events, reload configuration and checks
 	for range outChan {
+		// if the context has expired, then shut down the master status watcher entirely
+		select {
+		case <-ctx.Done():
+			log.Debugln("configReloader: stopped notifying config reloads due to context cancellation")
+			return
+		default:
+		}
+
 		err := cfg.Load(configPath)
 		if err != nil {
 			log.Errorln("configReloader: Error reloading config:", err)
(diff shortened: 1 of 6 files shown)


integrii:
Great find!  I feel like this one should go into a different place, though.

Check out the `control` code [here](https://github.com/kuberhealthy/kuberhealthy/blob/master/cmd/kuberhealthy/kuberhealthy.go#L218).  This control loop is responsible for doing various check starts and stops and reloads.  It feels like the `configReloader` func should be refactored to return a channel, then after reloading configuration, it should notify the channel that configuration has changed.  That signal would go into the switch in the `control` code loop and result in this reload (but only if master!).

This way we have a central place to manage all reloading and check start/stopping control flow...

integrii:
You could probably move the config reloader goroutine to be just above the control flow... up where there call to `monitorKHJobs()` is...

joshulyne:
> Great find! I feel like this one should go into a different place, though.
> 
> Check out the `control` code [here](https://github.com/kuberhealthy/kuberhealthy/blob/master/cmd/kuberhealthy/kuberhealthy.go#L218). This control loop is responsible for doing various check starts and stops and reloads. It feels like the `configReloader` func should be refactored to return a channel, then after reloading configuration, it should notify the channel that configuration has changed. That signal would go into the switch in the `control` code loop and result in this reload (but only if master!).
> 
> This way we have a central place to manage all reloading and check start/stopping control flow...

makes sense! will refactor to do so!

cmd/kuberhealthy/kuberhealthy.go
```diff
@@ -236,6 +243,14 @@ func (k *Kuberhealthy) Start(ctx context.Context) {
 				k.RestartChecks(ctx)
 				k.RestartReaper(ctx)
 			}
+		case <-configReloadChan:
+			log.Infoln("control: Witnessed a kuberhealthy configuration change...")
+
+			// if we are master, stop, reconfigure our khchecks, and start again with the new configuration
+			if isMaster {
+				log.Infoln("control: Reloading external check configurations due to kuberhealthy configuration update")
+				k.RestartChecks(ctx)
```
joshulyne, line 252: @integrii should we also be restarting the Reaper -- like in the case above? 
https://github.com/kuberhealthy/kuberhealthy/pull/942/files#diff-6fe2eb1bda3003e83c281ea43553e293915389e31314d523116afe24efe877e3R244

integrii: I would restart the reaper just in case.  This way we are more likely to make any new configuration options that have been set for it take effect.

cmd/kuberhealthy/config.go
```diff
@@ -154,8 +154,8 @@ func startConfigReloadMonitoringWithSmoothing(filePath string, scrapeInterval ti
 	return outChan, cancelFunc, nil
 }
 
-// configReloader watchers for events in file and restarts kuberhealhty checks
-func configReloader(ctx context.Context, kh *Kuberhealthy) {
+// configReloadNotifier watchers for events in file, reloads the configuration, and notifies upstream to restart checks
```
integrii, line 157: This is really good, but it looks like `configReloadNotifier` will run forever.  This means that every time `kuberhealthy.Start()` is called, a new go routine will be spawned that never shuts down.  That is a very uncommon thing, but not impossible.

can you add a context here, then pass in the context when this routine is spawned?  Then, check if the context has expired and `return` if it has.

integrii (changes requested):

jonnydawg (approved):
thanks!

PR #970, opened 2021-06-25
Reaper NPE bug - ensure container terminated state exists before fetching container terminated time

joshulyne:


integrii (approved):
LGtM

PR #973, opened 2021-07-01
set owner references only on pods in the kuberhealthy namespace

mikeinton:
Addresses https://github.com/kuberhealthy/kuberhealthy/issues/903


joshulyne (approved):
lgtm!

integrii:
This also included a fix for our end to end check.

PR #989, opened 2021-07-27
Allow helm values to overwrite the default loglevel

willem-jans:


integrii:
Thanks for the PR!  I half noticed this was spamming yesterday but assumed it was part of our normal debugging stuff.  I didn't realize this was going out as a default setting.  Could you change the default from `debug` to `error`?

willem-jans:
Will do, main reason for setting it to debug is because I didn't want to change default behavior.

integrii:
I totally understand.  Not changing project defaults is the right call when we haven't made an oops :-)

integrii (approved):

PR #1001, opened 2021-08-09
policy/v1beta1 downgrade from policy/v1 for PDB

### LINKED ISSUE #1000: Chart PDB does not work on Kubernetes Version <=1.20
**Describe the bug**
Currently the PDB hard codes the api version to policy/v1 which pdb is not part of that before 1.21
https://github.com/kuberhealthy/kuberhealthy/blob/master/deploy/helm/kuberhealthy/templates/poddisruptionbudget.yaml#L3

**Steps To Reproduce**
- Deploy a kubernetes 1.20 cluster
- Deploy kuberheathly with pdb enabled

**Expected behavior**
PDB is deployed

**Versions**
 - Cluster OS: CentOS 7
 - Kubernetes Version: 1.20
 - Kuberhealthy Release or build [e.g. 0.1.5 or 235] 76

**Additional context**
Would recommend allowing to template the PDB version or have two pdb template specific to each version.

integrii:
This should fix #1000 - `policy/v1beta1` will have support end in Kubernetes 1.25.  Moving to `policy/v1` will cause versions before 1.20 to stop working.

### DIFF
--- a/deploy/helm/kuberhealthy/templates/poddisruptionbudget.yaml
+++ b/deploy/helm/kuberhealthy/templates/poddisruptionbudget.yaml
@@ -1,6 +1,6 @@
 ---
 {{- if .Values.podDisruptionBudget.enabled }}
-apiVersion: policy/v1
+apiVersion: policy/v1beta1
 kind: PodDisruptionBudget
 metadata:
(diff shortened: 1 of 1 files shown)


jonnydawg (approved):
Thanks!

PR #1011, opened 2021-10-09
Signature refactor for better context propagation

raidancampbell:
This is a continuation of the request in #640 to propagate the context more thoroughly throughout the codebase.

I've made a few design decisions that I'd like to draw attention to, in case maintainers disagree:

1. if there exists no context in any caller function all the way up to the top of the callstack, and the top of the callstack is an exported function, a `TODO` comment along with a `context.TODO` is placed at the beginning of the exported function.  This is meant as a note for a future refactor where the exported function can receive a context
2. Same as 1, but if the top of the callstack is an HTTP request handler, the context is taken from the request and no `TODO` comment is left.
3. context is always the first parameter. [one existing context parameter](https://github.com/kuberhealthy/kuberhealthy/blob/master/cmd/daemonset-check/run_check.go#L103) was reorganized from this.

It's worth mentioning that I'm unable to test this change (or the current `master` branch it's based on) due to build failures in the test: [health.WorkloadDetails](https://github.com/kuberhealthy/kuberhealthy/blob/master/pkg/metrics/exporter_test.go#L72) is undefined, and a handful of tests appear to be making HTTP requests to my k8s cluster.  Given the nature of this code change, I don't expect my lack of positive testing to be problematic: the code should compile or fail.

### DIFF
--- a/cmd/resource-quota-check/resources.go
+++ b/cmd/resource-quota-check/resources.go
@@ -67,7 +67,7 @@ func runResourceQuotaCheck(ctx context.Context) {
 }
 
 // examineResourceQuotas looks at the resource quotas and makes reports on namespaces that meet or pass the threshold.
-func examineResourceQuotas(ctx2 context.Context, namespaceList *v1.NamespaceList) chan []string {
+func examineResourceQuotas(ctx context.Context, namespaceList *v1.NamespaceList) chan []string {
 	resultChan := make(chan []string)
 
 	resourceQuotasJobChan := make(chan *Job, len(namespaceList.Items))


integrii:
You did a ton of work here to make the program more durable - and its amazing.  Thank you for the contribution.

I have a couple thoughts:
- Some of the kuberhealthy packages may be used by other programs, requiring them to implement code changes.  However, go modules should help with this.
- This change was done in one PR across every binary and package.  It would have been nicer to tackle smaller bits of the codebase at once.  Some of these contexts that are passed don't really end up making a difference because they stem from the same parent global context or use a simple `context.Background()`.  

All that said, I did some build tests and this seems to work fine!  I am going to OK a merge to `master` here, but I want to run some release candidates built off master before this ends up in a release.  Again, this is a ton of work, and while not fully implemented in all the code touched, its about 700 lines better than what we have now!

Thanks for your PR, again.

raidancampbell:
Apologies for the ham-fisted approach here, I totally understand if you want this reorganized by package, or closed and reimplemented at your leisure.

As far as the intent of the change: I don't expect any behavioral impact from this.  As you mentioned, most of these callstacks ultimately derive their context from an exported function, where they're instantiated as `context.TODO()`, the global context in the `main` package is now only referenced for instantiation and propagated to `runResourceQuotaCheck`.

The goal of this was to resolve #640 as fully as possible without introducing any breaking changes in function signatures.  That way, the path is mostly paved for when you're ready to make a breaking change and add context as a parameter to those exported functions.

To the best of my knowledge, since only unexported functions were changed this shouldn't cause any breakages: I'll happily help fix anything though!

As far as it not being implemented everywhere: I wasn't able to find any remaining instances where a `context.Background()` or `context.TODO()` was abused.  I'd like to continue fixing it now rather than provide another partial solution for #640, but that's just my preference

integrii:
See if you can get the DCO bot to pass:

`Commit sha: fb4dfeb, Author: R. Aidan Campbell, Committer: R. Aidan Campbell; The sign-off is missing.`

integrii:
Thanks again!

PR #1021, opened 2022-01-17
Chart fix to use AppVersion by default, allow override via values.yaml

TBeijen:
Minor fix to the helm chart:

* Chart `values.yaml` incorrectly had old version `v2.4.1`
* Template ignored image tag in `values.yaml` so previous point didn't have actual effect

Now by default `Chart.AppVersion` is used while allowing to override the tag if needed.


integrii (approved):
Nice helm-fu here.

PR #1022, opened 2022-01-19
fix State.AddError() empty str check

zach593:


Zach593:
@integrii

integrii (approved):

integrii:
Nice catch.  Thanks @zach593 !

PR #1024, opened 2022-02-07
Fix helm template indentation for securityContext in storage check

### LINKED ISSUE #994: Storage check adding securityContext to resources instead of container in template
**Describe the bug**
Enabled the storage checker and the chart is failing to deploy:
```
error validating data: ValidationError(KuberhealthyCheck.spec.podSpec.containers[0].resources): unknown field "securityContext" in io.github.comcast.v1.KuberhealthyCheck.spec.podSpec.containers.resource
```

Looks like in the template the securityContext needs to be moved back a tab.
https://github.com/kuberhealthy/kuberhealthy/blob/master/deploy/helm/kuberhealthy/templates/khcheck-storage.yaml#L56

**Steps To Reproduce**
- deploy helm chart with storage check enabled and securityContext enabled

**Expected behavior**
storage check is applied

BadLiveware:
This fixes the issue described in #994 which makes the predefined storage check not work

### DIFF
--- a/deploy/helm/kuberhealthy/templates/khcheck-storage.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-storage.yaml
@@ -52,11 +52,11 @@ spec:
           memory: {{ $.Values.check.storage.resources.limits.memory }}
           {{- end }}
         {{- end }}
-        {{- if $.Values.securityContext.enabled }}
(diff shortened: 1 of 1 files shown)


integrii (approved):
Tests pass.  Thank you!

PR #1026, opened 2022-02-21
fix extraEnvs in values.yaml

hans-d:
some 'extraEnvs' don't have a default value, and are missing '{}'. This prevents helm from setting those values with a key/value.

### DIFF
--- a/deploy/helm/kuberhealthy/values.yaml
+++ b/deploy/helm/kuberhealthy/values.yaml
@@ -122,7 +122,7 @@ check:
       registry: kuberhealthy
       repository: daemonset-check
       tag: v3.3.0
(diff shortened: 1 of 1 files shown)


integrii:
This is due to the DCO bot requiring your `git` commits to be signed.  This is required by the CNCF to ensure that the committer's name can not be spoofed.

image

If you hit details on this, [it will tell you how to sign your commits](https://github.com/kuberhealthy/kuberhealthy/pull/1026/checks?check_run_id=5271756002).

integrii (approved):

integrii:
I set DCO to pass manually here.  Thanks for the quick contribution!

PR #1039, opened 2022-03-01
The correction for deploymen-check code

Hungrylion2019:


Hungrylion2019:
Result never be nil ,the result should be empty RequestResult{}
@integrii

integrii (approved):
I tweaked the error message for this, too.

Nice catch!  That channel is not returning a pointer and therefore would never hit that code.

integrii:
If trip.com is using Kuberhealthy, I would love to see that added to the [adopters list](https://github.com/kuberhealthy/kuberhealthy/blob/master/docs/KUBERHEALTHY_ADOPTERS.md)!

PR #1040, opened 2022-03-01
metrics exporter: fix formatting bug

bengadbois:
Removes trailing pipe character, which can cause errors for downstream metrics consumers, such as [the veneur metrics parser](https://github.com/stripe/veneur/blob/ed29417822e2ae8f2cc4b8f1bda37ab4e7c75768/samplers/parser.go#L396-L400)

### DIFF
--- a/pkg/metrics/exporter.go
+++ b/pkg/metrics/exporter.go
@@ -42,6 +42,7 @@ func GenerateMetrics(state health.State) string {
 			for _, error := range d.Errors {
 				errors += fmt.Sprintf("%s|", error)
 			}
+			errors = strings.TrimSuffix(errors, "|")
 		}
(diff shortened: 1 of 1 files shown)


integrii (approved):
LGTM

integrii:
Thanks!  If Square is using Kuberhealthy, we would love to have it added to the [adopters list](https://github.com/kuberhealthy/kuberhealthy/blob/master/docs/KUBERHEALTHY_ADOPTERS.md)!

ausmith:
@integrii That's a good question. The work @bengadbois has been accomplishing with this tool seems to be providing a lot of potential value that we're experimenting with now! When we are in production and happy with operations there, we'll endeavor to get review from internal teams for approval on a public endorsements in the adopters list you linked.

integrii:
I really appreciate that!

PR #1050, opened 2022-03-18
add service.go 'return' when service is nil

Hungrylion2019:


integrii (approved):
Great find as always!

PR #1051, opened 2022-03-28
modify the dns-checker yaml and readme

Hungrylion2019:

### DIFF
--- a/cmd/dns-resolution-check/README.md
+++ b/cmd/dns-resolution-check/README.md
@@ -44,9 +44,9 @@ spec:
 
 ### Checking endpoints behind DNS Service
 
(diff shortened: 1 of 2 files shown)


Hungrylion2019:
I notice that the code uses "DNS_POD_SELECTOR" to filter the dns pod but the yaml uses "DNS_NODE_SELECTOR"
This is clearly wrong for different descrpition between code and yaml.
There is no doubt that we should change them to keep the same.
I changed it and thought that the description of "DNS_POD_SELECTOR"  of the code was more reasonable than "DNS_NODE_SELECTOR" 
code:
```
	labelSelector = os.Getenv("DNS_POD_SELECTOR")
	if len(labelSelector) > 0 {
		log.Infoln("Looking for DNS pods with label:", labelSelector)
	}
```
@integrii

integrii (approved):
Agreed.  LGTM!  Thanks and good to see you again @Hungrylion2019.

PR #1059, opened 2022-05-06
fix: fix registry variable default values

dirien:
Hi,

This PR fixes the handling of the `registry` variable in the `values.yaml`.

During my installation of kuberhealthy at my workplace, I wanted to set our internal Container Registry via the global `imageRegistry`. But, when setted the chart rendered the deployment without the `kuberhealty` for the Docker Hub repository part:

```bash
 containers:
      - image: docker.io/kuberhealthy:v2.7.1
        command: [/app/kuberhealthy]
        ports:

```
So, i set now the `registry` to `docker.io` and moved the former registry value to the repository part.

```bash
 containers:
      - image: docker.io/kuberhealthy/kuberhealthy:v2.7.1
        command: [/app/kuberhealthy]
        ports:

```

Now it works fine :)

Looking for your feedback.

FYI: @Kaktor

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -21,3 +21,4 @@
 - [Yashvardhan Kukreja](https://twitter.com/yashkukreja98)
 - [Zachary Hanson](mailto:Zachary_Hanson@comcast.com)
 - [Zhicheng Li](https://github.com/Hungrylion2019)
+- [Engin Diri](https://twitter.com/_ediri)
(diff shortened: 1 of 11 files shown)


integrii:
I think this makes sense, but it is a significant change to our check URLs.  Right now, check URLs automatically assume `docker.io` if no docker registry URL is supplied. This change would make Kuberhealthy more compatible with different types of container runtimes that may not assume `docker.io` like Docker its self does...

I am not sure that the `"` (double quotes) made any difference here, but the registry changes probably did.  I think it looks well enough!  Lets give it a shot...

integrii (approved):

Issue #1054, opened 2022-04-06
DNS internal check does not have permission to access node information when using helm chart deployment


tompittson:
**Describe the bug**
The internal DNS check has an error in the logs trying to read node information on the cluster after deployment with the helm chart.

**Steps To Reproduce**
- Deploy kuberhealthy to a kubernetes cluster using the helm chart with dnsInternal.enabled: true
- Wait for the DNS status internal check to run then view the logs of the pod that executed the check. The following messages are in the logs:
```
time="2022-04-06T16:15:50Z" level=info msg="Check pod is running on node: ip-10-123-123-123.ec2.internal"
time="2022-04-06T16:15:50Z" level=debug msg="Pod is on node: ip-10-123-123-123.ec2.internal"
time="2022-04-06T16:15:50Z" level=error msg="Error waiting for node to reach minimum age: nodes \"ip-10-123-123-123.ec2.internal\" is forbidden: User \"system:serviceaccount:kuberhealthy:dns-internal-sa\" cannot get resource \"nodes\" in API group \"\" at the cluster scope"
```

**Expected behavior**
The service account used by the pod should have permission to read the nodes. This isn't working because the helm chart creates a Role and RoleBinding but nodes are cluster scoped so need a ClusterRole and ClusterRoleBinding.

Manually applying the ClusterRole and ClusterRoleBinding fixes the issue and the logs contain the following messages:
```
time="2022-04-06T17:03:52Z" level=info msg="Check pod is running on node: ip-10-123-123-123.ec2.internal"
time="2022-04-06T17:03:52Z" level=debug msg="Pod is on node: ip-10-123-123-123.ec2.internal"
time="2022-04-06T17:03:52Z" level=debug msg="Check running on node:  ip-10-123-123-123.ec2.internal with node age: 22h14m58.520731686s"
```

**Versions**
 - Kubernetes Version: 1.19.15
 - Kuberhealthy Release: v2.7.0
 - dns-resolution-check version: v.1.5.0
 - Kuberhealthy Helm Chart Release: 82

integrii:
Hrmm... did you get any errors when you applied your helm chart?  This RBAC seems to indicate the `get` permission for `nodes`: https://github.com/kuberhealthy/kuberhealthy/blob/master/deploy/helm/kuberhealthy/templates/clusterrole.yaml#L59

[...]

PR #1067, opened 2022-06-02
fix: change Role to ClusterRole for dns internal check

verdel:
If we use Role and RoleBinding then ServiceAccount does not have permissions on the Node object at the cluster scope

Because of this, an error appears in the logs of the pod:
```
Error waiting for node to reach minimum age: nodes "basic-8cpu-16gb-cefnd" is forbidden: User "system:serviceaccount:kuberhealthy:dns-internal-sa" cannot get resource "nodes" in API group "" at the cluster scope
```


mikeinton:
Related #1054

integrii (approved):
I think this works, but I am not sure that a `get` for all is the correct behavior once the scope is moved to the cluster level...  Should work, though.

PR #1068, opened 2022-06-02
Change check order to an empty value for runDuration in metrics exporter

verdel:
It seems to me that it would be more correct to first check `runDuration` for an empty value

If `runDuration` hasn't been set yet, ie. pod never run or failed to provision, set `runDuration` to 0 and only then parse `runDuration` (at this point `runDuration` will definitely not contain an empty string)


integrii (approved):
Agreed.  This makes sense.

Issue #1071, opened 2022-06-08
Custom Field with static value in the Report 


povils:
**Describe the feature you would like and why you want it**
Hey, it would be great if we somehow could set the custom fields with static values which would be reported in the health status output. Something like :
```json
    "OK": true,
    "Errors": [],
    "Cluster": "my-cluster-name" // custom field with static value
```

In our case, because we have many clusters and the routing to the correct cluster health endpoint might have bugs, it would be great if we could easily verfify that when we try to access my `/cluster/my-cluster-name/health` is indeed pointing and reporting the correct cluster health. I believe other static information could also be useful, like K8s version 

These custom fields can be as simple as provided as ENV variables to kuberhealhty: Example:
```yaml
KH_STATUS_FIELD_Cluster: "my-cluster-name"
KH_STATUS_FIELD_ClusterVersion: "1.24"
```
If the idea sounds good, I can implement

integrii:
This is an understandable feature request... This would go somewhere in [here](https://github.com/kuberhealthy/kuberhealthy/blob/06c6b9e45ac4186717d825f5e5e6415d1bfb04ff/cmd/kuberhealthy/kuberhealthy.go#L1548)...

povils:
@integrii I am taking this over then :)

PR #1073, opened 2022-06-12
Custom Metadata inside health state

povils:
Add custom Metadata to Health State (solves #1071)


integrii:
You did a lot of great stuff here, and thank you for that!  I feel like a prefixed environment variable might be a little messy and confusing to implement.  What do you think about implementing these variables instead as a map in the `configmap`?

The `configmap` is [here](https://github.com/kuberhealthy/kuberhealthy/blob/master/cmd/kuberhealthy/config.go#L14).  We could put a `map[string]string` in here that gets placed in the status page output if present.  That would avoid all the prefix parsing and the slightly confusing pattern of prefixing your metadata as environment variables.

Thoughts?

povils:
@integrii I think it is indeed better, I updated the PR :)

integrii (approved):
LGTM!  Sorry for the delay!  Thank you! 🎉

fungaren:
How about report by `/externalCheckStatus`:

https://github.com/kuberhealthy/kuberhealthy/blob/b489fb6fba14ec054fddc1cf18d0f29658cec6cd/pkg/checks/external/status/main.go#L1-L26

Should we support to add metadata there?

integrii:
Merged in #1158

PR #1078, opened 2022-06-16
adding ssh check to registry

rjacks161:
Updating documentation to include the new ssh check which will gather a list of nodes from the kube API, and attempt to ssh to them.


rjacks161:
I also don't have access to the kuberhealthy docker hub anymore, so that will need to be pushed if this is approved.

integrii:
I can request a paid docker account from the CNCF, which would let more users be added to the Kuberhealthy organization.  Until then, it's probably okay to push this to your own docker hub ID if you like.  Nothing in the project thus far declares that we have to push pages to the Kuberhealthy docker org...

This check looks really cool, though.  I am going to merge this one, but you might want to document the `SSH_EXCLUDE_LIST` environment variable.

Thanks a lot for this one. Good stuff!

integrii (approved):

PR #1082, opened 2022-08-16
Support for Image Pull Secret in Deployment Check containers

mckennajones:
Adds a new env var to the deployment check container, to specify `imagePullSecrets` in the deployment child pods. Useful if you have a locked down K8s cluster where you only want to pull images from private registries.


integrii (approved):
LGTM!  Thanks for the contribution!  Sorry for the delay here, but happy to have this little improvement 🎉

PR #1087, opened 2022-10-07
Add options to disable or reduce the errors label+values for Prometheus metrics

jkulesa:
Add a new configuration structure to kuberhealthy:

```
   promMetricsConfig:
      suppressErrorLabel: false 
      errorLabelMaxLength: 0   
```

... this is designed by default (if not used) to not change KuberHealthy prometheus behavior, but if defined, allows users to disable or limit the amount of prometheus string data that is included in the errors label value.  The rationale behind this is for customers that have prometheus setups with scrape_limit set, KH doesn't stop being scraped when too many errors are found, blowing up the scrape size, and preventing metrics publishing.

suppressErrorLabel = true (not the default) prevents the error label (and value) from being emitted at all, if set to false and 
errorLabelMaxLength is set and >0, then the error label value (per check) is limited to a number of bytes as configured.

Both these values give people some power and control when needing to limit prometheus endpoint error data to keep a robust KH environment running.

Then a bunch of small fixes:
* additions to metrics tests + fixes of existing metrics tests
* add metrics tests to kuberhealthy Dockerfile since they don't need external cluster validation
* adjust e2e a little to allow for unstable tag to be tested against easily [remove global helm since the chart doesn't use global namepace]
* Makefile - add .phony line for test (and other tgts)
* upgrade helm PDB to policy/v1 since k8s 1.25 its dropped entirely
*


integrii (approved):
I am not familiar with prometheus to understand the driver behind this change, but the code looks fine.  The `.Capabilities.APIVersions.Has` statement in the helm chart was new to me as well.  Very nice.

I appreciate you including tests and taking the time to explain your PR fully.

It would be nice to have the helm chart augmented to include this in the configmap, but that can be a future PR - this one is big enough.

LGTM 🎉 

Thanks!

PR #1088, opened 2022-10-11
Build multi-arch docker images for kuberhealthy(amd64/arm64)

code-haven:
Addresses: https://github.com/kuberhealthy/kuberhealthy/issues/166


cmd/kuberhealthy/Makefile
```diff
@@ -1,12 +1,11 @@
-IMAGE="kuberhealthy/kuberhealthy"
-TAG="unstable"
+include ../../Makefile.common.mk
+
+BUILDER := "kuberhealthy"
```
code-haven, line 3:
```suggestion
BUILDER := "dockerx-kuberhealthy"
```

code-haven:
Hey @integrii, I've added support to build kuberhealthy's docker images to support arm64 architecture by publishing multi-arch images using `docker buildx` Would really appreciate if you could given an initial review to get a sense if I'm going in the right direction. Thank you!

integrii (approved):
This is a really interesting implementation, but I appreciate how the main makefile continues to work as expected.  Thanks for maintaining that pattern instead of creating a giant monolithic makefile!

I am happy to get some ARM images pushed in the next release.

LGTM 🎉

PR #1089, opened 2022-10-15
Fix CWE G114 blocking builds

integrii:
this blocks all other PR's from being merged because they all fail this security check...

- import tidy
- disable gosec G114
- use go 1.18


jdowni000 (approved):
Lgtm

PR #1095, opened 2022-11-15
Fix scoping issue in storage check helm template

diversario:
Correctly refer to the top-level `imagePullSecrets` value.


integrii (approved):
Thanks for coming to contribute here!  I think it makes sense that we reference the global `imagePullSecret`, but I think it could be possible that different images in this chart could use different secrets in some rare situations.  It makes me think that each check should be its own subchart or something some day...

Thanks again!

Issue #1109, opened 2023-02-09
Allow restricting container syscalls for Kuberhealthy pods in Helm chart


erichstoekl:
**Describe the feature you would like and why you want it**
Cluster administrators may wish to take advantage of `seccompProfile` in order to restrict the syscalls that the kuberhealthy-spawned containers are able to make. Details of what `seccompProfile` does and how to enable it are documented here: https://kubernetes.io/docs/tutorials/security/seccomp/

For example, some clusters may enforce `seccompProfile` on all pods deployed to the cluster in order to meet security or compliance goals.

It would be helpful if the helm chart is capable of allowing for `seccompProfile` configuration.

# PR
PR is submitted for this issue as #1110

ketanjani21:
@erichstoekl - can this issue be closed , as https://github.com/kuberhealthy/kuberhealthy/pull/1110/files addresses it?

integrii:
Closing as resolved: seccompProfile support was added in PR #1110 (merged 2023-03-07). If you still see it missing in the chart/templates for a specific release, please comment or reopen with details.

PR #1110, opened 2023-02-09
Add seccompProfile parameter to the securityContext document for kuberhealthy pods

erichstoekl:
# Description
This PR is to address the issue: #1109 

In this PR we add the configuration option to enable `seccompProfile` for pods created by kuberhealthy. We do so by adding conditions to yaml templates, wherein if the `seccompProfile` object exists in `values.yaml`, it will be applied to the relevant templates.

# Testing
This change was tested on minikube by enabling all relevant `checks` with the `seccompProfile` object in place (set to `type: RuntimeDefault`). The Helm chart successfully applies and pods spawn normally.


integrii (approved):

integrii:
I have no ability to rebase this, so I am merging because I know it will pass upstream.

Issue #1055, opened 2022-04-18
Should client-gen be used to generate clients?


dmizelle:
**Describe the feature you would like and why you want it**
I'm attempting to write an operator that requires functionality of creating `KuberhealthyJob` objects and grabbing their `KuberhealthyState` objects after they run, and performing additional actions based on their output.

As part of writing this operator, I'd like to be able to mock out requests to a cluster so that I don't have to run one via `kind` or `envtest`. However, looking at the clients provided in https://github.com/kuberhealthy/kuberhealthy/blob/master/pkg/apis/khstate/v1/kuberhealthystate.go, it looks like that they are manually maintained. The copyright notice still refers to the "Knative maintainers."

From what I can see, we have the "annotations" for using client-gen (https://github.com/kuberhealthy/kuberhealthy/blob/master/pkg/apis/khstate/v1/types.go#L7), but we don't appear to be using client-gen anywhere.

Using client-gen would give users of this project:

* Automatically versioned clientsets that they can use to interact with Kuberhealthy resources on a cluster, removing the requirement for this project to maintain their own.
* Fake clients (a  la `k8s.io/client-go/kubernetes/fake`) so that they may mock out requests to a cluster themselves in testing.

**Additional context**
I can implement this functionality myself via PR, but I'd like to understand if there is a reason if the project has chosen not to use client-gen. Mainly so that I don't waste anyone's time :)

Thanks for the project!

dmizelle:
I found that I could mock this out better with `sigs.k8s.io/controller-runtime/pkg/client/fake` instead:

```go
func NewMock(objs ...runtime.Object) *Client {
        scheme := runtime.NewScheme()

        AddToScheme(scheme)

        client := fake.NewFakeClientWithScheme(scheme, objs...)

        return &Client{client}
}

func AddToScheme(scheme *runtime.Scheme) {
        gv := schema.GroupVersion{
                Group:   "comcast.github.io",
                Version: "v1",
        }
        scheme.AddKnownTypes(
                gv,
                &khcheck.KuberhealthyCheck{},
                &khcheck.KuberhealthyCheckList{},
        )

        scheme.AddKnownTypes(
                gv,
                &khjob.KuberhealthyJob{},
                &khjob.KuberhealthyJobList{},
        )

        scheme.AddKnownTypes(
                gv,
                &khstate.KuberhealthyState{},
                &khstate.KuberhealthyStateList{},
        )

[...]

PR #1125, opened 2023-05-02
Add status to khcheck and khjob

### LINKED ISSUE #1120: `KuberhealthyCheck` `status` field instead of separate `KuberhealthyState` object

Currently the status of a `KuberhealthyCheck` is stored in a separate `KuberhealthyState` object of the same name. Is there a good reason why this isn't instead stored in a `status` field of the `KuberhealthyCheck` itself, as is customary in Kubernetes?

**Additional context**

The separate object makes it impossible to create a custom Argo CD [health check](https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#custom-health-checks) for the `KuberhealthyCheck`, which would allow Argo CD to show the application as unhealthy if a check is failing.

ugurzongur:
Fixes #1120

After changing the `types.go` files for `khjob` and `khcheck`, I manually edited `pkg/apis/khjob/v1/kuberhealthyjob.go` and `pkg/apis/khcheck/v1/kuberhealthycheck.go` files respectively because I couldn't replicate the method that was used to generate those files. This is also mentioned in #1055. I tried current and older versions of https://github.com/kubernetes/code-generator with various parameters but they all ended up being quite different from the existing files especially in terms of folder structure. I believe those files are a combination of auto-generation and post-editing. IMHO that needs to be fixed eventually.

### DIFF
--- a/.ci/_job_test.sh
+++ b/.ci/_job_test.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+# ========== Utility functions ==========
+function get_khjob_status_ok {
+    kubectl get khjob -n $NS kh-test-job -ojsonpath='{.status.ok}'
+}
+
+function get_khstate_ok {
+    kubectl get khstate -n $NS kh-test-job -ojsonpath='{.spec.OK}'
+}
+
+function job_phase {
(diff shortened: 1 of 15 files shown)


LS80:
@ugurzongur How can we get this change merged?

LS80:
😞

PR #1126, opened 2023-05-04
Rename prometheusrule.yml to prometheusrule.yaml

kuritka:
I have a projects with many Helm charts that I am deploying in my infrastructure. For many practical reasons I use zero-trust `.gitignore` (I do allow-list instead of deny-list). Because of a probably forgotten typo I have to add a new rule to `.gitignore`. This PR fixes it.

Signed-off-by: Michal Kuritka <kuritka@gmail.com>


integrii:
Sounds fine!

PR #1136, opened 2023-07-12
Add namespace value to override globally where to deploy kuberhealthy.

cdesaintleger:
This PR allow to override the namespace where to deploy kuberhealthy. 
Usefull as exemple when you want to deploy it as a subchart in a different namespace.


integrii:
This looks fine to me!  Thank you for doing this.

PR #1137, opened 2023-07-14
Add support for Istio

kalamburajay:
Service check needs to be adapted to envoy proxy http response


kalamburajay:
Is the project still active because i see no traction on existing PRs for review

integrii:
The project has not been given love and maintenance from me for some time, but I am picking it back up.  My employer also is very particular about any side projects taking away time from the company.

PR #1141, opened 2023-08-18
Re-add listenNamespace configmap value

jsvk:
Kuberhealthy was made namespace-capable by this PR: https://github.com/kuberhealthy/kuberhealthy/pull/461

This allowed us to deploy namespaced instances of Kuberhealthy which would co-exist with each other on the cluster, which fit well within our multitenancy model - different orgs and teams could deploy (and be billed for, and responsible for the uptime of) their own Kuberhealthy instances.

As part of the subsequent move to the config file-based configuration, this configuration value was dropped, causing Kuberhealthy to always watch the entire cluster.

This PR re-adds support by properly unmarshalling and passing the configuration to the right watch calls. 

The CONFIGURATION.md still references the value although it didn't work: https://github.com/kuberhealthy/kuberhealthy/blob/master/docs/CONFIGURATION.md


integrii:
Thank you!  I appreciate the quick `buildx` change as well.

PR #1147, opened 2023-10-30
fix(http-check): remove password logging with basic auth

AloysAqemia:
It is possible to configure basic auth credentials by passing them directly in the URL with this format: `http://user:password@domain.tld/path`

With such a configuration, `http-check` logs the password in many places.

This patch uses the `url.Redacted()` method to suppress the password (but not the username) from the logs and the generated metrics.


integrii:
Thank you for making the http check more secure!

PR #1166, opened 2023-11-28
executing a request to the kh endpoint instead of creating a Request object

elebertus:
This addresses https://github.com/kuberhealthy/kuberhealthy/issues/1165

### DIFF
--- a/pkg/checks/external/nodeCheck/main.go
+++ b/pkg/checks/external/nodeCheck/main.go
@@ -85,15 +85,17 @@ func waitForKuberhealthyEndpointReady(ctx context.Context, kuberhealthyEndpoint 
 			return doneChan
 		default:
 		}
-
-		_, err := http.NewRequest("GET", kuberhealthyEndpoint, nil)
-		if err == nil {
+		res, err := http.Get(kuberhealthyEndpoint)
+		if err != nil {
+			log.Debugln(kuberhealthyEndpoint, "Error making reuqest to kuberhealthy endpoint "+err.Error())
+			time.Sleep(time.Second * 3)
+			continue
+		}
+		if res.StatusCode == 200 {
 			log.Debugln(kuberhealthyEndpoint, "is ready.")
 			doneChan <- nil
 			return doneChan
-		} else {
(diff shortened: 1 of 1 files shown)


elebertus:
Rebased for signoff to satisfy the DCO bot.

pkg/checks/external/nodeCheck/main.go
```diff
@@ -85,15 +85,17 @@ func waitForKuberhealthyEndpointReady(ctx context.Context, kuberhealthyEndpoint
 			return doneChan
 		default:
 		}
-
-		_, err := http.NewRequest("GET", kuberhealthyEndpoint, nil)
-		if err == nil {
+		res, err := http.Get(kuberhealthyEndpoint)
+		if err != nil {
+			log.Debugln(kuberhealthyEndpoint, "Error making reuqest to kuberhealthy endpoint "+err.Error())
```
thatfatguypat, line 90: You've got a typo here on `reuqest` , but otherwise I think this looks sensible

elebertus: Got it, thanks!

pkg/checks/external/nodeCheck/main.go
```diff
@@ -79,21 +79,25 @@ func waitForKuberhealthyEndpointReady(ctx context.Context, kuberhealthyEndpoint
 	doneChan := make(chan error, 1)
 
 	for {
+		res, err := http.Get(kuberhealthyEndpoint)
+		if err != nil {
+			log.Debugln(kuberhealthyEndpoint, "Error making request to kuberhealthy endpoint "+err.Error())
+			continue
+		}
+		if res.StatusCode == 200 {
```
martinsmatthews, line 87: possibly we don't need a 200, just getting a response/err being nil is enough?

elebertus: I think technically we could still get a non-200 response and a non-nil error. So checking that we get an healthy response is appropriate still.

elebertus:
I’ll update the go mod bits to try and get the build to pass 👍

PR #1244, opened 2024-04-13
Update Grafana Dashboard

peschmae:
- Replaces the deprecated Graph & Table Panels, with the new Timeseries & Table panels
- Make the Datasource an input variable, that can be chosen during import of the dashboard


integrii:
Thanks!  I will take this for now even though I don't have a good way to test it at the moment.  I will be revisiting this as I work on the next KH release, which is actively going on...

Issue #1165, opened 2023-11-28
nodeCheck.waitForKuberhealthy doesn't actually make a request to kuberhealthy


martinsmatthews:
**Describe the bug**

The comment above [nodeCheck.waitForKuberhealthy](https://github.com/kuberhealthy/kuberhealthy/blob/master/pkg/checks/external/nodeCheck/main.go#L77-L99) says `waitForKuberhealthyEndpointReady hits the kuberhealthy endpoint every 3 seconds to see if the node is ready to reach the endpoint.` but in fact that function only builds the request, it doesn't then use it with `http.Client.Do(..)`. 

This may be deliberate, but then the description doesn't really make sense to me.

**Steps To Reproduce**

Doesn't really need reproduction, it can be seen in the code.

**Expected behavior**

I would expect it to create the http.Client on line 80, perhaps with a 1 second timeout, and then, after creating the request, add a call to execute the request
```
_, err = client.Do(req)
```
and then just check for the error, if there is no error then return success/push to the doneChan.

I'm happy to PR a fix for this, if you agree this is a bug.

**Screenshots**
If applicable, add screenshots to help explain your problem.  You can paste images into github and they will upload automatically.

**Versions**
 - Cluster OS: [e.g. CoreOS 1800.1.0]
 - Kubernetes Version: 1.9.7
 - Kuberhealthy Release or build [e.g. 0.1.5 or 235]

**Additional context**

elebertus:
Doing some quick testing in a [fork](https://github.com/kuberhealthy/kuberhealthy/compare/master...elebertus:kuberhealthy:master)  by changing the call to a plain `http.Get` I was able to get this working very quickly.

I do see the same issue @martinsmatthews posted, since just an `http.Request` is being created an http request is never made to the kuberhealthy endpoint. So in reality all that check is doing is testing that `net/http` can created a value `http.Request` object with the given parameters.

ketanjani21:
Submitted PR https://github.com/kuberhealthy/kuberhealthy/issues/1165 to address this.

integrii:
Closing as resolved: PRs #1259 and #1308 merged to ensure nodeCheck.waitForKuberhealthy actually executes the HTTP request. Please reopen if this persists in the latest release.

PR #1259, opened 2024-05-28
fix http request made to kh healthcheck endpoint

nak0f:
Fixes same issue in #1165


integrii (approved):
Thank you and sorry for the long delay on approval!  This will go out when the next release is finally debugged.

PR #1276, opened 2024-09-02
fix: add probes in e2e deploy check

aso-adeo:
I think that the cleanup done here https://github.com/kuberhealthy/kuberhealthy/commit/a15f56457de6a537d2de02526357d242202b8b23, in order to be compliant with the new k8s.go/api 1.29 version, needed a replacement to be able to start the deployment-deployment container with a livenessProbe and a readinessProbe set.


integrii (changes requested):
Thanks for the contribution!  Welcome to the project. I'm not 100% sure, but it looks like `tcpSocket` is unset, so this code will explode.  Is that right?

xavierbaude (approved):
LGTM

PR #1286, opened 2024-11-05
feat: Allow extraLabels and extraAnnotations to be configured per check

maxforasteiro:
## Description

Hey there!

We have been dealing with some errors in kuberhealthy due to the concurrent nature of kuberhealthy and karpenter, one fighting to consume resources and the other fighting to set them free. 

One possible way to deal with this is to not allow karpenter to disrupt kuberhealthy checks with annotation the annotation `karpenter.sh/do-not-evict: "true"`

According to my ~deep~ investigation, [both extraAnnotations and extraLabels are already supported by khchecks](https://github.com/kuberhealthy/kuberhealthy/blob/master/pkg/apis/khcheck/v1/types.go#L37-L39) so we just need to put them in the values and pass it to the checks 🚀

I'm not sure about the `version` change as I would always have a semver instead of a single integer but I'm not here to fix the world, right?

I rendered the chart locally and everything seems to work.

### DIFF
--- a/deploy/helm/kuberhealthy/templates/khcheck-daemonset.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-daemonset.yaml
@@ -8,6 +8,14 @@ metadata:
 spec:
   runInterval:  {{ .Values.check.daemonset.runInterval }}
   timeout: {{ .Values.check.daemonset.timeout }}
+  {{- if .Values.check.daemonset.extraAnnotations }}
+  extraAnnotations:
+    {{- toYaml .Values.check.daemonset.extraAnnotations | nindent 4 }}
+  {{- end }}
+  {{- if .Values.check.daemonset.extraLabels }}
+  extraLabels:
+    {{- toYaml .Values.check.daemonset.extraLabels | nindent 4 }}
+  {{- end }}
   podSpec:
     {{- if .Values.check.daemonset.serviceAccountName }}
     serviceAccountName: {{ .Values.check.daemonset.serviceAccountName }}
--- a/deploy/helm/kuberhealthy/templates/khcheck-daemonset.yaml
+++ b/deploy/helm/kuberhealthy/templates/khcheck-daemonset.yaml
@@ -8,14 +8,8 @@ metadata:
 spec:
   runInterval:  {{ .Values.check.daemonset.runInterval }}
   timeout: {{ .Values.check.daemonset.timeout }}
-  {{- if .Values.check.daemonset.extraAnnotations }}
-  extraAnnotations:
-    {{- toYaml .Values.check.daemonset.extraAnnotations | nindent 4 }}
-  {{- end }}
-  {{- if .Values.check.daemonset.extraLabels }}
-  extraLabels:
-    {{- toYaml .Values.check.daemonset.extraLabels | nindent 4 }}
-  {{- end }}
+  {{- include "kuberhealthy.extraAnnotations" .Values.check.daemonset.extraAnnotations | nindent 2 }}
+  {{- include "kuberhealthy.extraLabels" .Values.check.daemonset.extraLabels | nindent 2 }}
   podSpec:
     {{- if .Values.check.daemonset.serviceAccountName }}
     serviceAccountName: {{ .Values.check.daemonset.serviceAccountName }}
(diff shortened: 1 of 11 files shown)


deploy/helm/kuberhealthy/templates/khcheck-daemonset.yaml
```diff
@@ -8,6 +8,14 @@ metadata:
 spec:
   runInterval:  {{ .Values.check.daemonset.runInterval }}
   timeout: {{ .Values.check.daemonset.timeout }}
+  {{- if .Values.check.daemonset.extraAnnotations }}
+  extraAnnotations:
```
ivankatliarchuk, line 12: Could we have instead a helper function instead of direct use of the value?

The main benefit, is we could override behaviour for extra annotations and extra labels

https://github.com/kuberhealthy/kuberhealthy/blob/master/deploy/helm/kuberhealthy/templates/_helpers.tpl

{{- define "kuberhealthy.extra-labels" -}}
{{- define "kuberhealthy.extra-annotations" -}}

maxforasteiro: hm, I don't think it would work, since these values are set per predefined checks (daemonset, deployment, dns-external, etc) and I would prefer to be able to set these per check and not per chart (unless I'm missing something obvious here)

I could implement a `globalExtraAnnotations` and `globalExtraLabels` but that was not my goal with this PR.

maxforasteiro: @ivankatliarchuk do you agree with my comment? how do you see this being implemented?

ivankatliarchuk: And another point. 
Example

```tpl
 {{- if .Values.check.daemonset.extraAnnotations }}
  extraAnnotations:
    {{- toYaml .Values.check.daemonset.extraAnnotations | nindent 4 }}
  {{- end }}
```

This means annotations need to be set per check, when is nice to have some sort of global annotations

```
  extraAnnotations:
    {{- toYaml .Values.check.annotations | nindent 4 }}
    {{- toYaml .Values.check.daemonset.extraAnnotations | nindent 4 }}   
```

Not sure if this supported

maxforasteiro: @ivankatliarchuk okay, templating was implemented, but not the global annotations. I bumped into a problem while trying to implement the global annotation which would make the solution not clean.

```tpl
{{- define "kuberhealthy.extraAnnotations" }}
  {{- if . -}}
  extraAnnotations:
    {{- toYaml .Values.check.extraAnnotations | nindent 2 }}
    {{- toYaml . | nindent 2 }}
  {{- end -}}
{{ end -}}
```

^ this wouldn't work because the template doesn't have the full context (`.Values`) loaded on it and passing it as an extra arg to all templating calls would create the same problem that we were trying to avoid with the other method. I could add if you see the value on it 😄

maxforasteiro: @ivankatliarchuk I would love to have your feedback here. My kuberhealthy checks keep failing with false-positives because of karpenter killing the nodes before the kuberhealthy checks are done and adding the annotation would fix it

maxforasteiro: @integrii @jandersen-plaid @nak0f sorry for tagging you, but it has been hard to communicate with @ivankatliarchuk about this issue

maxforasteiro: @ivankatliarchuk Have you checked the last commit? I've implemented the templating; it's just that I couldn't achieve what you suggested.

maxforasteiro:
hey @ivankatliarchuk can you reopen the PR? I can't from my end

maxforasteiro:
tagging @integrii and @sheikhrachel also since @ivankatliarchuk has been really unresponsive

ivankatliarchuk:
Hi @maxforasteiro . I can't reopen a PR. 

Please note that I'm not a maintainer of this repository. Just someone following this repository and is also a user for kuberhealthy.

PR #1288, opened 2024-11-26
fix: securityContext for some built-in checks

acunabit:
This pull request introduces the following updates to the Kuberhealthy Helm chart:

- Support for securityContext.capabilities. Allows fine-grained control over container capabilities to improve security configurations.
- Addition of runAsNonRoot to the DaemonSet and DNS External Checks. runAsNonRoot is available in all the other checks.


chitha20:
i still don't see support for this, do u know why this was closed and not released ?

PR #1291, opened 2024-12-19
Fixes for node checks

nak0f:
PR adds below fixes in Kuberhealthy readiness checks:
- Check http response code in requests made to Kuberhealthy endpoint. This is for catching issues in where there are intermediate proxies between checker and Kuberhealthy, like an Istio sidecar proxy.
- Use health check path (/) in WaitForKuberhealthy() instead of reporting path. Reporting path (/externalCheckStatus) responds with http 400 to readiness checks as it expects report payloads from KHchecks.
- Introduce a default http timeout in waitForKuberhealthyEndpointReady() in order to prevent http requests blocking the check longer than the context timeout.
- Add negative test to cover failure cases mentioned above

### DIFF
--- a/pkg/checks/external/nodeCheck/main.go
+++ b/pkg/checks/external/nodeCheck/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"net/url"
 	"os"
 	"time"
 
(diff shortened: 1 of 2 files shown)


nak0f:
@integrii : Could you please take a look when available?

PR #1293, opened 2025-02-24
feat: Allow extraLabels and extraAnnotations to be configured per check

maxforasteiro:
## Description

Hey there!

We have been dealing with some errors in kuberhealthy due to the concurrent nature of kuberhealthy and karpenter, one fighting to consume resources and the other fighting to set them free. 

One possible way to deal with this is to not allow karpenter to disrupt kuberhealthy checks with annotation the annotation `karpenter.sh/do-not-evict: "true"`

According to my ~deep~ investigation, [both extraAnnotations and extraLabels are already supported by khchecks](https://github.com/kuberhealthy/kuberhealthy/blob/master/pkg/apis/khcheck/v1/types.go#L37-L39) so we just need to put them in the values and pass it to the checks 🚀

I'm not sure about the `version` change as I would always have a semver instead of a single integer but I'm not here to fix the world, right?

I rendered the chart locally and everything seems to work.

### DIFF
--- a/deploy/helm/kuberhealthy/Chart.yaml
+++ b/deploy/helm/kuberhealthy/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: kuberhealthy
 appVersion: v2.8.0-rc3
-version: 105
+version: 106
 home: https://kuberhealthy.github.io/kuberhealthy/
 description: "An operator for synthetic test and monitoring. Works great with Prometheus."
 type: application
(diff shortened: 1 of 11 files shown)


integrii (approved):
This LGTM!  Thank you!

In the future, all checks will be in their own repos, so this chart might be blown apart.  Until then though, this is good to go.

maxforasteiro:
hey @integrii when do you plan to have this released?

integrii:
This one actually goes out to the Helm chart automatically with an action: https://github.com/kuberhealthy/kuberhealthy/actions/workflows/flat-files-generation.yml

PR #1295, opened 2025-03-06
Add basic auth option for image pull check for private repositories

ghitabot:
Hello kuberhealthy team!
I have an on premise use case where I want to use Kuberhealthy checks to monitor various parts of the infrastructure setup. One problem that I'm facing is that the image pull checker doesn't allow configuration for private repositories. I know that in cloud setups you can assign identities and gain access to container registry service, but for on premise setups it's easier to just provide the credentials to the check.
I prepared this small PR to include a flag and 2 new parameters to specifiy if you want to set basic auth login to the target registry and to provide the username / password.

Thanks.
Ghita Bot.

### DIFF
--- a/cmd/image-download-check/main.go
+++ b/cmd/image-download-check/main.go
@@ -1,7 +1,9 @@
 package main
 
 import (
+	"github.com/google/go-containerregistry/pkg/authn"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/crane"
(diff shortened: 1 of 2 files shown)


cmd/image-download-check/main.go
```diff
@@ -36,6 +47,11 @@ func main() {
 
 	var err error
 
+	// set login to true if the LOGIN_REQUIRED env var is set to true
+	if strings.ToLower(os.Getenv("LOGIN_REQUIRED")) == "true" {
```
integrii, line 51: could do a `ParseBool` here, but this is fine I suppose

integrii (approved):

integrii:
Thanks for the PR.  KH is currently getting a total rewrite so that it builds off of the standard patterns of `kubebuilder`. I'm happy to merge this as-is though for now. In the future, all checks will be in their own repositories so releases will be easier.

PR #1483, opened 2025-10-06
Trim namespace whitespace when reading service account namespace

integrii:
## Summary
- trim whitespace when reading the service account namespace file and allow the path to be overridden for tests
- add unit tests covering namespace trimming and the default fallback behavior

## Testing
- go test -count=1 ./...

------
https://chatgpt.com/codex/tasks/task_e_68e32391b8b883238e693f502c1ed967

### DIFF
--- a/cmd/kuberhealthy/util.go
+++ b/cmd/kuberhealthy/util.go
@@ -2,9 +2,14 @@ package main
 
 import (
 	"os"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


integrii:
@codex do not use parallel testing on TestGetMyNamespaceTrimsWhitespace or any other test that uses a global value.

PR #1494, opened 2026-03-13
Fix error propagation in deployment check

nak0f:
PR fixes error propagation issues in the deployment check where timeout and context cancellation were not reported back to the caller. 

In [service_requester.go](https://github.com/kuberhealthy/kuberhealthy/blob/master/cmd/deployment-check/service_requester.go#L73-L79) the `case <- timeoutChan` and `case <- ctx.Done()` branches do not return the actual errors. 

The caller: [run_check.go](https://github.com/kuberhealthy/kuberhealthy/blob/master/cmd/deployment-check/run_check.go#L141) only reacts to non-nil errors, so timeouts and context cancellation are ignored and main flow continues as nothing happened.


integrii:
Thank you, but the master branch is about to be locked and switched over to the main branch as part of the upcoming V3 release!

PR #1500, opened 2026-04-02
fix: prevent duplicate checker pod creation on startup

### LINKED ISSUE #1499: Duplicate health check pods happen just after bootstrap
**Describe the bug**
Just after when Kuberhealthy is setup, the deployment healthcheck fails for the first instance because two pods for the healthcheck get created at the same time. 

**Steps To Reproduce**
- install Kuberhealthy with the healthchecks together
- The deployment healthcheck fails for the first instance of the check because of duplicate k8s service deployment-svc

**Expected behavior**
Healthchecks should pass in the first attempt if the cluster is already healthy

**Versions**
 - Kubernetes Version: 1.34.2

kushwiz:
Add a per-check mutex guard in StartCheck to prevent the informer handleAdd and the scheduler loop from racing and creating two pods for the same health check simultaneously.

A tryStartCheck/finishStartCheck pattern ensures only one goroutine can enter StartCheck for a given HealthCheck at a time.

Fixes #1499

### DIFF
--- a/internal/kuberhealthy/kuberhealthy.go
+++ b/internal/kuberhealthy/kuberhealthy.go
@@ -81,6 +81,9 @@ type Kuberhealthy struct {
 
 	metricsMu       sync.Mutex
 	metricsSnapshot controllerMetricsSnapshot
+
+	checkStartMu sync.Mutex                    // guards checkStarting map access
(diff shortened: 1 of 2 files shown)


integrii:
Fixed with improvements and other corrections in #1503! Thank you.

cooperlees:
Sweet - thanks!

PR #1511, opened 2026-06-10
Fix/release start block if check reported to replica

Awlie:
Hi @integrii,

Great work on the huge refactor!

I started using v3 yesterday and noticed that, after a few scheduled runs, most of my checks no longer started. I cloned the Helm files and kept `replicas: 2`.

After analyzing which checks start and which do not start after restarting the Kuberhealthy deployment, I found that the issue is caused by the in-memory start guard not being released on the leader pod if the check response is received by a non-leader. Without releasing it, it keeps failing with `check {check-name} is already being started`.

I tested this change in my cluster, and now the lock is successfully released and all checks start as expected.

```log
time="2026-06-10T11:07:02Z" level=info msg="modified checker pod" name=http-internal namespace=kuberhealthy
time="2026-06-10T11:07:02Z" level=debug msg="releasing the in-memory start flag" name=http-internal namespace=kuberhealthy
time="2026-06-10T11:07:02Z" level=info msg="KHCheck resource updated" name=http-internal namespace=kuberhealthy pod=http-internal
```

But maybe you have a better spot where this could fit?

Thanks again for your work on this!

Best regards  
Oliver

#### Go package versions

When looking at the Go package versions, a `3.0.1` version exists that is actually older than `3.0.0`. See: https://pkg.go.dev/github.com/kuberhealthy/kuberhealthy/v3/pkg/api?tab=versions  

#### Rewriting of `runInterval` and `timeout` in CRD

 I am using two Helm deployments: one for Kuberhealthy (deployed first), and another for my checks and a custom 
orchestrator, which reads the CRDs to report and transform results to a different backend. Kuberhealthy seems to rewrite the `runInterval` and `timeout` values from, for example, `5m` to `5m0s`.  
  Because of this, I had to adjust the values in my second Helm deployment to match the updated CRD syntax. Otherwise, I would get:

```txt
conflict occurred while applying object kuberhealthy/dns-status-internal kuberhealthy.github.io/v2, Kind=HealthCheck: Apply failed with 2 conflicts: conflicts with "kuberhealthy" using kuberhealthy.github.io/v2:
- .spec.runInterval
- .spec.timeout
```

and many more. After adjusting the values to include `0s`, the Helm upgrades worked without issues.

#### edited this PR

originally i have done these changes aswell, but i have skipped them for now, also the scope was too big.

I also updated the Go version, dependencies, and the GitHub Actions to match the newer Go version and due to the following warning

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

If that’s too many changes, I’m happy to revert the build-related updates. I can only test the github action changes in this PR.

### DIFF
--- a/internal/kuberhealthy/kuberhealthy.go
+++ b/internal/kuberhealthy/kuberhealthy.go
@@ -1107,8 +1107,8 @@ func (kh *Kuberhealthy) UpdateCheck(oldKHCheck *khapi.HealthCheck, newKHCheck *k
 		"namespace": newKHCheck.Namespace,
 		"name":      newKHCheck.Name,
 		"pod":       newKHCheck.Name,
-	}).Info("KHCheck resoruce updated")
-	// TODO - do we do anything on updates to reload the latest check? How do we prevent locking into an infinite udpate window with the controller?
+	}).Info("KHCheck resource updated")
+	// TODO - do we do anything on updates to reload the latest check? How do we prevent locking into an infinite update window with the controller?
 
 	// // stop the check
 	// err := kh.StopCheck(oldKHCheck)
--- a/internal/kuberhealthy/kuberhealthy.go
+++ b/internal/kuberhealthy/kuberhealthy.go
@@ -542,7 +542,8 @@ func (kh *Kuberhealthy) handleCreate(ctx context.Context, khc *khapi.HealthCheck
 	}
 }
 
-// handleUpdate reconciles finalizers and restarts checks when their spec changes.
+// handleUpdate reconciles finalizers, restarts checks when their spec changes,
+// and clears in-memory start flags when UUID is cleared (across instances).
 func (kh *Kuberhealthy) handleUpdate(ctx context.Context, oldCheck *khapi.HealthCheck, newCheck *khapi.HealthCheck) {
 	if !newCheck.GetDeletionTimestamp().IsZero() {
 		if kh.hasFinalizer(newCheck) {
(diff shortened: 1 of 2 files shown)


Awlie:
the test weren't failing before, i need to take a look at it. 

somehow this one test is now flaky, sometimes the test passes, sometimes it doesn't.

also the engineerd/setup-kind github action update didn't work for now, i'll look for a solution in the next days

Awlie:
i cherry picked 7a9003ce into this branch and then rebased it on top of main

i hope everything looks good now.

i will create a separate issue or merge request for github actions and go dependencies and the other findings

integrii:
Hey Awlie. I am sorry about that bot spam here. It's a bit of an... Experiment. 

I appreciate your replies and contributions. I am investigating deeper into this thread and will send a better reply tomorrow.

integrii:
LGTM Thank you Awlie!

